PR #13, opened 2016-10-11
Retry to Fire Trigger

### LINKED ISSUE #4: Retry vs abort for triggers
When firing a trigger, there are some clear error cases where we want to simply retry firing the trigger. For example, when the user is throttled by firing triggers too quickly, the return code is `429`. In this case, it is sufficient to pause and retry firing the trigger.

However, other error conditions warrant completely stopping/deleting the consumer entirely. For example if a `404` is hit when firing the trigger, it is clear that the trigger has been deleted and the consumer should be stopped entirely.

There are certainly other possible errors when firing triggers, which should each be handled appropriately with either a retry or complete abort.

dubeejw:
- If a trigger does not fire, retry to fire it if the service was unavailable (503), and interal server error occurred (500), the request timed out (408), or the request was throttled (429)

Closes #4

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -17,6 +17,7 @@ import logging
 import requests
 import ssl
 import threading
+import time
 from database import Database
 from kafka import KafkaConsumer, KafkaProducer
 from kafka.errors import KafkaError
@@ -24,13 +25,14 @@ from kafka.structs import OffsetAndMetadata
 
 
 class Consumer (threading.Thread):
+    retry_timeout = 1   # Timeout in seconds
+    max_retries = 10    # Maximum number of times to retry firing trigger
 
     def __init__(self, trigger, params):
         threading.Thread.__init__(self)
 
         self.daemon = True
         self.shouldRun = True
-
         self.trigger = trigger
         self.isMessageHub = params["isMessageHub"]
         self.triggerURL = params["triggerURL"]
@@ -47,6 +49,7 @@ class Consumer (threading.Thread):
             self.parseAsJson = False
 
     def run(self):
+
         # TODO may need different code paths for Message Hub vs generic Kafka
         if self.isMessageHub:
             sasl_mechanism = 'PLAIN'       # <-- changed from 'SASL_PLAINTEXT'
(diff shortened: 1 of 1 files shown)


dubeejw:
@bjustin-ibm, how do you feel about this one?

provider/consumer.py
```diff
@@ -40,6 +41,8 @@ def __init__(self, trigger, params):
         self.password = params["password"]
 
     def run(self):
+        retry_timeout = 1
```
bjustin-ibm: Needs a comment for unit

provider/consumer.py
```diff
@@ -96,32 +99,32 @@ def run(self):
...
+                            retry = False
+                        elif status_code != 503 or status_code != 500 or status_code != 408 or status_code != 429:
+                            logging.error('Error talking to OpenWhisk, status code {}'.format(status_code))
+                            retry = False
+
+                    except requests.exceptions.RequestException as e:
+                        logging.error('Error talking to OpenWhisk: {}'.format(e))
+
+                    if retry:
+                        logging.info("Failed to commit the offset. Retrying in {} second(s)".format(retry_timeout))
```
bjustin-ibm: This message isn't exactly correct because you didn't even attempt to commit the offset. Maybe just limit it to saying "Retrying in {} second(s)". Also please prefix all these new logging messages with the trigger name as `"[{}] Retrying in {} second(s)".format(self.trigger, retry_timeout)`

provider/consumer.py
```diff
@@ -96,32 +99,32 @@ def run(self):
...
-
-                # TODO
-                # error handling
-                # retries? I suppose certain status codes warrant it...
-                # only commit offset if post was successful?
+
+                logging.info("Firing trigger {} with {} messages with a total size of {} bytes".format(self.trigger,
+                   len(messages), messageSize))
+
+                while retry:
```
bjustin-ibm: I don't see where `retry` is set to `True`

dubeejw:
Think I got everything.

provider/consumer.py
```diff
@@ -96,32 +99,33 @@ def run(self):
...
+                        # Manually commit offset if the trigger was fired successfully. Retry firing the trigger if the
+                        # service was unavailable (503), an internal server error occurred (500), the request timed out
+                        # (408), or the request was throttled (429).
+                        if status_code == 200:
+                            self.consumer.commit()
+                            retry = False
+                        elif status_code != 503 and status_code != 500 and status_code != 408 and status_code != 429:
+                            logging.error('Error talking to OpenWhisk, status code {}'.format(status_code))
+                            retry = False
+                            self.shouldRun = False
```
bjustin-ibm: I just realized that while the consume will stop, it will not be deleted from the DB. The next time this app fires up, the (broken) consumer will be recreated.

I think think can easily be fixed by moving the DB delete code from the `shutdown` method the outside of the main `while` loop.

dubeejw: I moved the mentioned code into the case where the trigger doesn't exist. Not sure if you want to leave it in shutdown() as well.

sjfink: if the user is over quota, they will get 429's right?    Don't think we want to retry in that case.

sjfink: also to protect us from willies, I think we should retry at most N times for some small N.

bjustin-ibm: @sjfink I was under the impression that the trigger quota was time-based and that backing off for a second or two will result in success. At least, I believe I have had some anecdotal evidence to suggest that.

provider/consumer.py
```diff
@@ -96,32 +99,35 @@ def run(self):
...
+                            database = Database()
+                            database.deleteTrigger(self.trigger)
```
bjustin-ibm: @dubeejw This needs to get moved to after `self.consumer.close()` (currently line 134) just outside the main `while` loop so that the delete also happens as a result of calling `shutdown()`

provider/consumer.py
sjfink: regarding the quota -- the situation I have in mind is a user who has not put a credit card on file and has exhausted their monthly free tier.     All  further usage (in the month) will be rejected.   Actually I don't know if we decided what code they get -- it may not be 429.   Anyway, in this case, we don't want to retry.

Regarding bounded retries -- I'm just scared of things running out of control if there's a bug anywhere.     Safety first, right?   I don't want a bug or failure to result in a DoS attack -- I'd rather drop events.   Note that I don't intend to "drop the feed", just drop events for the duration of the error condition rather than having huge amounts of work back up.

Another question: what if the kafka feed is just too heavy for OW to ever catch up?   e.g. if the kafka feed is generating 1 million messages per second and OW can consume only 1000, we don't want millions or billions of messages to back up, I think.

sjfink: yes, that sounds right.

PR #15, opened 2016-10-12
introduce isJSONData parameter

### LINKED ISSUE #11: special optional handling for JSON formatted messages
Although kafka message content is, strictly speaking, always a byte array, it seems that JSON formatted strings will be so common that it may warrant special handling.

It has been proposed that if the trigger creator specifies that the message format is a JSON formatted string (using a new `isJSONData` parameter to the feed action), then the feed service should make a reasonable attempt at parsing the message content and passing it appropriately as the trigger payload.

If `isJSONData` is set to true and the message cannot be parsed as valid JSON, the result is the same as if `isJSONData` were set to false - which is the current behavior. In this case, the feed service should log a warning message.

@csantanapr

bjustin-ibm:
This allows you to specify at trigger creation time if you expect the message value to be stringified JSON. If the parameter is true and the message value can be parsed as JSON, then the resulting JSON object will be part of the payload for the trigger - otherwise, the un-parsed message content will be sent.

Resolves #11

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -39,6 +39,13 @@ class Consumer (threading.Thread):
         self.username = params["username"]
         self.password = params["password"]
 
+        # handle the case where there may be existing triggers that do not
+        # have the isJSONData field set
+        if "isJSONData" in params:
+            self.parseAsJson = params["isJSONData"]
+        else:
+            self.parseAsJson = False
+
     def run(self):
         # TODO may need different code paths for Message Hub vs generic Kafka
         if self.isMessageHub:
@@ -84,7 +91,7 @@ class Consumer (threading.Thread):
                     logging.debug("Consumed message: {}".format(str(message)))
                     messageSize += len(message.value)
                     fieldsToSend = {
-                        'value': message.value,
+                        'value': self.parseMessageIfNeeded(message.value),
                         'topic': message.topic,
                         'partition': message.partition,
                         'offset': message.offset,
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw please review.

bjustin-ibm:
Trigger payload with `isJSONData -> false`:
```json
{
    "messages": [
        {
            "timestamp": 1476293299094,
            "timestamp_type": 0,
            "partition": 0,
            "key": null,
            "offset": 7623,
            "topic": "mytopic",
            "value": "{\"message\": \"At the sound of the beep, it is now 1476293299\"}"
        }
    ]
}
```

Triger payload for the same message with `isJSONData -> true`:
```json
{
    "messages": [
        {
            "timestamp": 1476293299094,
            "timestamp_type": 0,
            "partition": 0,
            "key": null,
            "offset": 7623,
            "topic": "mytopic",
            "value": {
                "message": "At the sound of the beep, it is now 1476293299"
            }
        }
    ]
}
```

provider/consumer.py
```diff
@@ -84,7 +91,7 @@ def run(self):
                     logging.debug("Consumed message: {}".format(str(message)))
                     messageSize += len(message.value)
                     fieldsToSend = {
-                        'value': message.value,
+                        'value': self.parseMessage(message.value),
```
bjustin-ibm: The function returns the right thing depending on whether the message should and can be parsed, and this is nice and easy to read. Perhaps instead I'll change the name to `parseMessageIfNeeded`

provider/consumer.py
```diff
@@ -134,3 +141,17 @@ def shutdown(self):
...
+            try:
+                parsed = json.loads(value)
+                logging.info('[{}] Successfully parsed a message as JSON.'.format(self.trigger))
+                return parsed
+            except:
+                # no big deal, just return the original value
+                logging.warn('[{}] I was asked to parse a message as JSON, but I failed.'.format(self.trigger))
+                pass
+
+        logging.info('[{}] Returning un-parsed message'.format(self.trigger))
```
bjustin-ibm: Then it wouldn't return the unparsed value if `self.parseAsJson` were `False`.

Issue #7, opened 2016-10-11
/health endpoint


bjustin-ibm:
Implement an endpoint that returns data regarding the health of the feed service provider. Information should include things like:
- average CPU load (past hour? array of averages for each of the past 24 hours?)
- memory usage
- number of active consumers

The caller of this endpoint is interested in determining how much work is being done by the feed service, and how much more work can it do before it is overloaded.

bjustin-ibm:
Other statistical information is also valuable (e.g. average time to process incoming messages, size of messages processed, etc.) but these might be better handled by statistical analysis of the log messages, rather than actively computing these inside the feed service provider itself.

bjustin-ibm:
I suspect this will need to be implemented as a separate thread that wakes up every now and then, computes and records the required information.

dubeejw:
@bjustin-ibm, how do you feel about having a separate thread running that periodically outputs the health information to the logs?

bjustin-ibm:
@dubeejw I suspect that will be necessary in order to provide "average" stats. However, it is possible that the `/health` endpoint could simply return the current values, and some other monitoring system will handle computing averages by periodically polling the `/health` endpoint and recording the results.

For now, let's just make it return the instantaneous values.  Later, if we have a need, we can compute/record values using a separate thread.

PR #22, opened 2016-10-19
Create /health route

bjustin-ibm:
Currently only contains the uptime in seconds

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+from datetime import datetime
 from flask import Flask, jsonify, request
 from consumer import Consumer
 from database import Database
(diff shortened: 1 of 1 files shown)


provider/app.py
```diff
@@ -65,6 +66,13 @@ def testRoute():
     return jsonify('pong')
 
 
+@app.route('/health')
+def healthRoute():
+    currentTime = datetime.now()
+    diff = currentTime - startTime
+    return jsonify({'uptime': diff.seconds})
```
dubeejw: I think at some point we could add more information to the JSON that is returned here. Possibilities include: how many triggers have been invoked, how many triggers failed to be invoked, how many consumers are there, etc.

dubeejw: Also, maybe a day(s), minute(s), second(s), hour(s) time format could useful here.

bjustin-ibm: Yup, agreed. There are more stats we could include that are documented in #7. For now I just want to get the uptime implemented to ensure we understand how Docker maintains state over time.

I'll switch the `uptime` to include days, hours, minutes, and seconds.

PR #23, opened 2016-10-19
return promise in feed action

bjustin-ibm:
The feed action does not actually return the Promise created in the `doRequest` function. This causes the feed action to always immediately "succeed" even when things go horribly wrong.

### DIFF
--- a/action/feedAction.js
+++ b/action/feedAction.js
@@ -80,6 +80,8 @@ function doRequest(options) {
             }
         });
     });
+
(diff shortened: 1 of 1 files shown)


action/feedAction.js
```diff
@@ -80,6 +80,8 @@ function doRequest(options) {
             }
         });
     });
+
+    return promise;
```
dubeejw: LGTM

PR #27, opened 2016-10-25
Keep track of each consumer's last successful poll

bjustin-ibm:
This will be reported in the `/health` endpoint as seconds elapsed since each consumer's last successful poll. Presumably, a high value will indicate that a consumer has died, a negative value indicates the consumer has not fully initialized.

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -89,10 +89,24 @@ def testRoute():
 
 @app.route('/health')
 def healthRoute():
+    healthReport = {}
+
     currentTime = datetime.now()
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw Please review.

provider/app.py
```diff
@@ -89,10 +89,24 @@ def testRoute():
 
 @app.route('/health')
 def healthRoute():
+    healthReport = {}
```
dubeejw: LGTM

PR #28, opened 2016-10-25
Do Not Create Duplicate Consumers

dubeejw:
- Reject requests to create consumers for the same trigger

Closes https://github.com/openwhisk/openwhisk-package-kafka/issues/26

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -33,30 +33,39 @@ consumers = dict()
 def postTrigger(namespace, trigger):
     body = request.get_json(force=True, silent=True)
     triggerFQN = '/' + namespace + '/' + trigger
-    logging.info("[{}] Ensuring user has access rights to post a trigger".format(triggerFQN))
 
-    trigger_get_response = requests.get(body["triggerURL"])
-    trigger_get_status_code = trigger_get_response.status_code
-    logging.info("[{}] Repsonse status code from trigger authorization {}".format(triggerFQN, trigger_get_status_code))
-
-    if trigger_get_status_code == 200:
-        logging.info("[{}] User authenticated. About to create consumer {}".format(triggerFQN, str(body)))
-        createAndRunConsumer(triggerFQN, body)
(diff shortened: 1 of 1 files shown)


dubeejw:
@bjustin-ibm, ready for review.

provider/app.py
```diff
@@ -31,27 +31,36 @@
...
-    elif trigger_get_status_code == 401:
-        logging.info("[{}] User not authorized to post trigger".format(triggerFQN))
+
+    if triggerFQN in consumers:
+        logging.info("[{}] Trigger already exists".format(triggerFQN))
+        trigger_get_status_code = 403
         result = {
             'success': False,
-            'error': 'not authorized'
+            'error': "already exists"
```
bjustin-ibm: "trigger already exists"

provider/app.py
```diff
@@ -31,27 +31,36 @@
...
-        logging.info("[{}] User authenticated. About to create consumer {}".format(triggerFQN, str(body)))
-        createAndRunConsumer(triggerFQN, body)
-        logging.info("[{}] Finished creating consumer.".format(triggerFQN))
-        result = {'success': True}
-    elif trigger_get_status_code == 401:
-        logging.info("[{}] User not authorized to post trigger".format(triggerFQN))
+
+    if triggerFQN in consumers:
+        logging.info("[{}] Trigger already exists".format(triggerFQN))
+        trigger_get_status_code = 403
```
dubeejw: @bjustin-ibm, why do you think 400 is better to use than the unauthorized error code?

bjustin-ibm: @dubeejw I suppose there is some room for interpretation here, but I prefer to interpret this as a bad request on the client because they asked to do something that does not need to be done. To me, this is an invalid request, rather than one they are not allowed to perform. Maybe another way to say it is that this exact request would likely be allowed if the trigger didn't already exist, but 403 says "you can't do this, don't ever try again."

dubeejw: Updated the error code to 409.

PR #29, opened 2016-10-26
use a "real" WSGI server

bjustin-ibm:
The server included with Flask is generally considered a "toy" that should not be used for production apps. Instead, let's use the WSGI server in gevent.


dubeejw:
@bjustin-ibm, LGTM.

PR #30, opened 2016-10-26
Remove Username and Password Parameters from Feed

dubeejw:
- Username and password should not be parameters for Kafka in feedAction.js


bjustin-ibm:
@dubeejw ready?

dubeejw:
@bjustin-ibm, go for it.

PR #33, opened 2016-10-28
Split feedAction into messageHubFeed and kafkaFeed

dubeejw:
- Create two different actions one for MessageHub, and one for Kafka


bjustin-ibm:
These actions share a lot in common. It may be worth investigating how to put that commonality into a shared file and using the new .zip action feature.

PR #34, opened 2016-10-28
Build and deploy

bjustin-ibm:
Supersedes #31 
Resolves #6 

Thanks for all your help @jasonpet!
- add a default route so that ansible deploy can make sure the app is running
- automatically create the Cloudant table if it doesn't yet exist
- log to file

PR #37, opened 2016-10-31
create database if it doesn't already exist

bjustin-ibm:


provider/database.py
```diff
@@ -18,13 +18,19 @@
 from cloudant.result import Result
 
 class Database:
+    dbname = "ow_kafka_triggers"
```
dubeejw: Starting to the database name should be passed in as a parameter to the Docker run command instead of being hardcoded.

bjustin-ibm: @dubeejw I see something similar in the cloudant and alarms trigger deployment scripts. I've ask Jason about it in #34, and we can handle it there if needed.

PR #39, opened 2016-10-31
Change the /ping route to just /

bjustin-ibm:
This can then be used by the deployment/monitoring scripts to ensure the server is running

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -91,9 +91,9 @@ def deleteTrigger(namespace, trigger):
     return response
 
 
-@app.route('/ping')
+@app.route('/')
 def testRoute():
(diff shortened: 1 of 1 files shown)


provider/app.py
```diff
@@ -82,7 +82,7 @@ def deleteTrigger(namespace, trigger):
     return response
 
 
-@app.route('/ping')
+@app.route('/')
 def testRoute():
     return jsonify('pong')
```
dubeejw: Not sure if "pong" is the appropriate message to display since you are removing "ping" from the URL.

bjustin-ibm: Is there something else you would like this endpoint to say? 😉

PR #41, opened 2016-11-01
Use package_endpoint parameter

### LINKED ISSUE #40: Parameterize service URL in feed actions
Currently, the feed actions have a hardcoded URL for the feed service provider. This should be parameterized so that this URL can be set during deployment.

bjustin-ibm:
This parameter should be set to the correct value as a bound parameter when the feed action is deployed.

Resolves #40

### DIFF
--- a/action/kafkaFeed.js
+++ b/action/kafkaFeed.js
@@ -8,11 +8,16 @@ var request = require('request');
  *  @param {string} endpoint - address to OpenWhisk deployment
  */
 function main(params) {
+    if(!params.package_endpoint) {
+        whisk.error('Could not find the package_endpoint parameter.');
+        return;
+    }
+
     var triggerComponents = params.triggerName.split("/");
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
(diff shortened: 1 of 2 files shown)


action/feedAction.js
```diff
@@ -14,7 +14,7 @@ function main(params) {
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
-    var feedServiceURL = 'http://owkafkafeedprovider.mybluemix.net/triggers/' + namespace + '/' + trigger;
+    var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
```
dubeejw: Add a check for the package_endpoint. Should be good after that.

bjustin-ibm:
This will need to be reworked due to #33

bjustin-ibm:
@dubeejw Please have one more look now that I've applied it to the two new actions.

action/kafkaFeed.js
```diff
@@ -8,11 +8,16 @@ var request = require('request');
  *  @param {string} endpoint - address to OpenWhisk deployment
  */
 function main(params) {
+    if(!params.package_endpoint) {
```
dubeejw: LGTM

PR #49, opened 2016-11-09
Check Message Hub credentials before creating feed

bjustin-ibm:
Use the Message Hub REST API to list existing topics - this verifies the credentials are correct
While we're at it, make sure the topic they want to subscribe to already exists.

### DIFF
--- a/action/messageHubFeed.js
+++ b/action/messageHubFeed.js
@@ -74,7 +74,6 @@ function checkMessageHubCredentials(params) {
     var options = {
         method: 'GET',
         url: topicURL,
-        json: true,
         headers: {
             'X-Auth-Token': params.api_key
         }
--- a/action/messageHubFeed.js
+++ b/action/messageHubFeed.js
@@ -29,20 +29,27 @@ function main(params) {
             return;
         }
 
-        var body = validatedParams;
-        body.triggerURL = 'https://' + whisk.getAuthKey() + "@" + params.endpoint + '/api/v1/namespaces/' + namespace + '/triggers/' + trigger;
+        // make sure we have valid MH credentials
+        console.log("Checking Message Hub credentials...");
+        return checkMessageHubCredentials(validatedParams)
+            .then(function() {
+                console.log("Successfully authenticated with Message Hub");
 
-        var options = {
-            method: 'PUT',
-            url: feedServiceURL,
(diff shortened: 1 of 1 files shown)


action/messageHubFeed.js
```diff
@@ -14,7 +14,7 @@ function main(params) {
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
-    var feedServiceURL = 'http://169.46.145.194:5000/triggers/' + namespace + '/' + trigger;
+    var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
```
bjustin-ibm: It is expected that this parameter will be bound when the action is installed. See #34.

action/messageHubFeed.js
```diff
@@ -55,6 +62,35 @@ function main(params) {
     }
 }
 
+function checkMessageHubCredentials(params) {
+    // listing topics seems to be the simplest way to check auth
+    var topicURL = params.kafka_admin_url + '/admin/topics';
```
bjustin-ibm: This is a new parameter that needs to be supplied when creating the trigger. It is part of the Bluemix credentials and will be "magically" present when using `package refresh`.

action/messageHubFeed.js
```diff
@@ -120,6 +156,20 @@ function validateParameters(rawParams) {
         return;
     }
 
+    if(rawParams.kafka_admin_url) {
+        validatedParams.kafka_admin_url = rawParams.kafka_admin_url;
```
bjustin-ibm: This is a new parameter that needs to be supplied when creating the trigger. It is part of the Bluemix credentials and will be "magically" present when using `package refresh`.

dubeejw: Update line 119 to hand the case where the body is not valid JSON. Besides that it works for me.

bjustin-ibm: @dubeejw The problem was that I was using both `json:true` in the request options _and_ calling `JSON.parse()` on the response body. In this way, `JSON.parse()` is trying to parse an already-parsed JSON object - not a string like it is expecting.

This should be fixed now.

dubeejw:
LGTM.

Issue #51, opened 2016-11-10
Remove Invalid Consumers from Dictionary 


dubeejw:
Consumers that have invalidated Message Hub credential, have a trigger that is deleted, or fail to invoke a trigger more than ten times must be removed from the main dictionary of consumers. Otherwise, these consumers will exist until the Docker container is restarted.

bjustin-ibm:
Resolved by #54

Issue #43, opened 2016-11-01
restart consumers when time since last poll is too high


bjustin-ibm:
I have confirmed an instance where a single consumer stopped polling due to issues communicating with Kafka _and_ the last successful poll time was significantly greater than the regular poll interval. This proves that a single consumer crash can be detected with this measurement.

Knowing this, we now need to gracefully handle this scenario and restart the consumer to recover from the error.

bjustin-ibm:
Resolved by #54

Issue #35, opened 2016-10-31
observed consumer crash


bjustin-ibm:
It appears there was either a network glitch or the kafka instance was unavailable during the poll. That being said, our code should handle this situation more gracefully.

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/KafkaFeedProvider/consumer.py", line 89, in run
    partition = self.consumer.poll(1000)
  File "/usr/local/lib/python2.7/site-packages/kafka/consumer/group.py", line 514, in poll
    records = self._poll_once(remaining)
  File "/usr/local/lib/python2.7/site-packages/kafka/consumer/group.py", line 541, in _poll_once
    self._coordinator.ensure_coordinator_known()
  File "/usr/local/lib/python2.7/site-packages/kafka/coordinator/base.py", line 221, in ensure_coordinator_known
    self._client.poll(future=metadata_update)
  File "/usr/local/lib/python2.7/site-packages/kafka/client_async.py", line 483, in poll
    self._maybe_connect(node_id)
  File "/usr/local/lib/python2.7/site-packages/kafka/client_async.py", line 322, in _maybe_connect
    conn.connect()
  File "/usr/local/lib/python2.7/site-packages/kafka/conn.py", line 304, in connect
    if self._try_handshake():
  File "/usr/local/lib/python2.7/site-packages/kafka/conn.py", line 366, in _try_handshake
    self._sock.do_handshake()
  File "/usr/local/lib/python2.7/ssl.py", line 830, in do_handshake
    self._sslobj.do_handshake()
error: [Errno 110] Operation timed out
```

bjustin-ibm:
On the plus side, this crash was detected using the code in #27 👍

bjustin-ibm:
Resolved by #54

PR #54, opened 2016-11-14
Actively monitor consumers - attempt to heal sick ones individually

### LINKED ISSUE #46: consumers do not recover from a network outage
I happened to be running the feed service on my laptop when the network completely went out (I pulled the plug). As you would expect, the one consumer I had started freaking out at not being able to connect to Message Hub.  However, it also did not recover once the network connection was restored.

The following messages were repeated indefinitely until I restarted the entire service:
```
WARNING:kafka.client:Node 1 connection failed -- refreshing metadata
ERROR:kafka.coordinator:Error sending GroupCoordinatorRequest_v0 to node 1 [NodeNotReadyError: 1]
WARNING:kafka.coordinator:Coordinator unknown during heartbeat -- will retry
WARNING:kafka.coordinator:Heartbeat failed ([Error 15] GroupCoordinatorNotAvailableError: ); retrying
```

bjustin-ibm:
Implement the "doctor" thread which monitors all consumers and attempts to heal their sickness by restarting them.

This is a pretty big change...

Consumers no longer _are_ Threads, but rather contain a ConsumerThread field. This allows the Consumer to separately monitor and replace the Thread doing the actual work. The Doctor wakes up every so often, monitors each of the Consumers and checks their health (is their actual state consistent with their desired state). If a Consumer is found to be sick, the Doctor attempts to restart that Consumer individually.

### DIFF
--- a/provider/thedoctor.py
+++ b/provider/thedoctor.py
@@ -1,0 +1,36 @@
+import logging
+import time
+
+from consumer import Consumer
+from consumercollection import ConsumerCollection
+from threading import Thread
+
+class TheDoctor (Thread):
+    sleepy_time = 2
+
+    def __init__(self, consumerCollection):
+        Thread.__init__(self)
+
+        self.daemon = True
+        self.consumerCollection = consumerCollection
+
+    def run(self):
+        logging.info('[Doctor] The Doctor is in!')
+
+        while True:
+            consumers = self.consumerCollection.getCopyForRead()
+
+            for consumerId in consumers:
+                consumer = consumers[consumerId]
+                logging.info('[Doctor] [{}] Consumer is in state: {}'.format(consumerId, consumer.currentState()))
+
+                if consumer.currentState() is Consumer.State.Dead and consumer.desiredState() is Consumer.State.Running:
+                    # well this is unexpected...
+                    logging.error('[Doctor][{}] Consumer is dead, but should be alive!'.format(consumerId))
+                    consumer.restart()
+                elif consumer.currentState() is Consumer.State.Dead and consumer.desiredState() is Consumer.State.Dead:
+                    # Bring out yer dead...
+                    logging.info('[{}] Removing dead consumer from the collection.'.format(consumer.trigger))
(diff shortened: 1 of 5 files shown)


dubeejw:
ERROR:root:[/.../messageHubTrigger] Uncaught exception: class State has no attribute 'Restarted'

provider/thedoctor.py
```diff
@@ -0,0 +1,35 @@
...
+        logging.info('[Doctor] The Doctor is in!')
+
+        while True:
+            consumers = self.consumerCollection.getCopyForRead()
+
+            for consumerId in consumers:
+                consumer = consumers[consumerId]
+                logging.debug('[Doctor] [{}] Consumer is in state: {}'.format(consumerId, consumer.currentState()))
+
+                if consumer.currentState() is Consumer.State.Dead and consumer.desiredState() is Consumer.State.Running:
```
dubeejw: This seems to work fix the scrolling socket disconnect issue:
```
elif consumer.secondsSinceLastPoll() > 2 and consumer.desiredState() is Consumer.State.Running:
    logging.error('[Doctor][{}] Consumer timed-out, but should be alive!'.format(consumerId))
    consumer.restart()
```

bjustin-ibm:
> class State has no attribute 'Restarted'

@dubeejw nice catch. This has since been fixed.

provider/thedoctor.py
bjustin-ibm: This makes me both happy and sad 🎭  Thanks.

bjustin-ibm: 846e05b583a3aff422b504e9a6101ba9d4a7efcc

bjustin-ibm: Added in 5517fbdeb569c64c6dfe28fa09d29a82ca452a6f

bjustin-ibm:
Resolves #46, #35, #51, and #43

bjustin-ibm:
What do you think @dubeejw? Merge it?

dubeejw:
@bjustin-ibm, merge it.

PR #57, opened 2016-11-15
Do not Assume all Fields are Provided when a Trigger is Posted

dubeejw:
- Ensure all needed fields are received from the client when posting a trigger
- Return an error JSON payload if proper fields are not received
- Don't error in Consumer if isMessageHub is set to false and credentials are not provided

Closes https://github.com/openwhisk/openwhisk-package-kafka/issues/56

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+import sys
 from datetime import datetime
 from flask import Flask, jsonify, request
 from consumer import Consumer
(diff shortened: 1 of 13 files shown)


provider/app.py
```diff
@@ -141,6 +149,24 @@ def restoreTriggers():
...
+    missing = []
+    required = ['brokers', 'topic', 'triggerURL', 'isMessageHub']
+
+    if fields is None:
+        missing = required
+    else:
+        missing.extend([i for i in required if i not in  fields])
+
+        if 'isMessageHub' in fields and fields['isMessageHub'] == True:
+            missing.extend([i for i in ['username', 'password'] if i not in fields])
```
bjustin-ibm: Instead of checking the params twice, how about optionally adding `username` and `password` to the `required` array, and then only checking once?

dubeejw: For Kafka, the username and password arguments are not required though.

PR #63, opened 2016-11-16
Increase poll timeout

### LINKED ISSUE #62: Increase poll timeout limit
I am experiencing a situation where the time to post a trigger to OW is taking more than 2 seconds. Since this time contributes to the entire `poll()` time, simply posting a trigger results in a poll timeout and consumer restart.

Increase the timeout to something more reasonable.

bjustin-ibm:
Increase poll timeout.

Resolves #62

### DIFF
--- a/provider/thedoctor.py
+++ b/provider/thedoctor.py
@@ -7,7 +7,7 @@ from threading import Thread
 
 class TheDoctor (Thread):
     # maximum time to allow a consumer to not successfully poll() before restarting
-    poll_timeout_seconds = 2
+    poll_timeout_seconds = 20
 
     # interval between the Doctor making rounds
     sleepy_time_seconds = 2


provider/thedoctor.py
```diff
@@ -7,7 +7,7 @@
 
 class TheDoctor (Thread):
     # maximum time to allow a consumer to not successfully poll() before restarting
-    poll_timeout_seconds = 2
+    poll_timeout_seconds = 20
```
dubeejw: How did you decide on 20 seconds?

bjustin-ibm: I noticed (anecdotally) that posting to a trigger URL can sometimes take over 5 seconds. The timeout value should be greater than that, plus allow any time for actually processing/parsing messages. Do you have some other value in mind?

PR #65, opened 2016-11-17
Do not Assume all Fields are Provided when a Trigger is Posted

dubeejw:
- Ensure all needed fields are received from the client when posting a trigger
- Return an error JSON payload if proper fields are not received
- Don't error in Consumer if isMessageHub is set to false and credentials are not providedi

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -41,6 +41,7 @@ def postTrigger(namespace, trigger):
     body = request.get_json(force=True, silent=True)
     triggerFQN = '/' + namespace + '/' + trigger
     expectedRoute = urllib.quote('/namespaces/' + namespace + '/triggers/' + trigger)
+    missing = getMissingPostFields(body)
 
     if consumers.hasConsumerForTrigger(triggerFQN):
         logging.warn("[{}] Trigger already exists".format(triggerFQN))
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw I added a commit here (22045bcf56dde3ccb9602989a8a1dbcf7faf61e9) that simplifies this a little bit. Please review.

provider/app.py
```diff
@@ -152,16 +152,16 @@ def restoreTriggers():
 
 def getMissingPostFields(fields):
     missing = []
-    messageHubRequired = ['brokers', 'topic', 'triggerURL', 'isMessageHub', 'username', 'password']
-    kafkaRequired = ['brokers', 'topic', 'triggerURL', 'isMessageHub']
+    requiredFields = ['brokers', 'topic', 'triggerURL', 'isMessageHub']
+
+    # Message Hub also requires 'username' and 'password fields'
+    if 'isMessageHub' in fields and fields['isMessageHub'] == True:
```
dubeejw: What if fields is None here, does an exception occur?

bjustin-ibm: Good point. Fixed in 0f85b6600cf6be8488c80b055a2ee56579d2cef9.

provider/app.py
```diff
@@ -154,13 +154,13 @@ def getMissingPostFields(fields):
...
 
-    # Message Hub also requires 'username' and 'password fields'
-    if 'isMessageHub' in fields and fields['isMessageHub'] == True:
-        requiredFields.extend(['username', 'password'])
-
     if fields is None:
         missing.extend(requiredFields)
     else:
+        # Message Hub also requires 'username' and 'password fields'
+        if 'isMessageHub' in fields and fields['isMessageHub'] == True:
```
dubeejw: LGTM

Issue #55, opened 2016-11-15
Consumer becomes unresponsive without exception - socket disconnected


bjustin-ibm:
It has been observed that a Consumer will get into a state where it is indefinitely repeating "socket disconnected" messages to the log. In the meantime, it is not consuming new messages.

This problem seems to come from the kafka-python client itself, where (I suppose) it is getting into an error-recovery loop.

bjustin-ibm:
Perhaps related to https://github.com/dpkp/kafka-python/issues/849 but that issue only mentions this from a producer side, as well as claiming that functionality shouldn't be affected (just annoying log messages).

bjustin-ibm:
Resolved by #67

PR #67, opened 2016-11-28
Confluent client

bjustin-ibm:
Resolve stability issues by switching to the confluent-kafka client.

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -16,13 +16,14 @@ import json
 import logging
 import requests
 import ssl
-from threading import Thread, Lock
 import time
+
+# HEADS UP! I'm importing confluent_kafka.Consumer as KafkaConsumer to avoid a
+# naming conflict with my own Consumer class
+from confluent_kafka import Consumer as KafkaConsumer, KafkaError
 from database import Database
 from datetime import datetime
-from kafka import KafkaConsumer, KafkaProducer
-from kafka.errors import KafkaError
-from kafka.structs import OffsetAndMetadata
+from threading import Thread, Lock
 
 
 class Consumer:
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw I think this is ready for review.

bjustin-ibm:
Resolves #55

provider/consumer.py
```diff
@@ -210,73 +198,66 @@ def run(self):
         except Exception as e:
             logging.error('[{}] Uncaught exception: {}'.format(self.trigger, e))
         finally:
+            logging.info('[{}] Cleaning up consumer'.format(self.trigger))
+            logging.debug('[{}] Closing KafkaConsumer'.format(self.trigger))
+            self.consumer.unsubscribe()
```
dubeejw: Will unsubscribe() or close() throw exceptions?

provider/consumer.py
```diff
@@ -210,73 +198,66 @@ def run(self):
...
+                # append Message Hub specific config
+                config.update({'ssl.ca.location': '/etc/ssl/certs/',
+                                'sasl.mechanisms': 'PLAIN',
+                                'sasl.username': self.username,
+                                'sasl.password': self.password,
+                                'security.protocol': 'sasl_ssl'
+                             })
+
+            consumer = KafkaConsumer(config)
+            consumer.subscribe([self.topic])
```
dubeejw: Does subscribe throw an exception?

provider/consumer.py
bjustin-ibm: @dubeejw This is addressed in c024c451827c69d1ffb54459c1ac67ee5cab5db6. Because the previous try/catch catches all exceptions, the cleanup code can be moved outside of the `finally` block and it will behave the same. This allows any possible exceptions to be caught, without the need for nested try/catch blocks.

PR #70, opened 2016-11-30
Package refresh

bjustin-ibm:
Create/update the package with the `bluemixServiceName` parameter set to `messagehub` so that package refresh works.


bjustin-ibm:
Depends on #34. Once that PR is merged, most of the changes (currently) attached to this PR will just get rebased away.

csantanapr:
LGTM 👍

PR #73, opened 2016-12-01
Generate a random UUID for consumers

bjustin-ibm:
This is used to present triggers in the /health endpoint without showing the user's namepace. Each consumer will have a random UUID to present instead of the trigger FQN.

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -14,7 +14,11 @@
 
 import logging
 import os
+import requests
 import sys
+import urllib
+import uuid
+
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


bjustin-ibm:
@dubeejw @csantanapr Please review.

csantanapr:
from design point, I can say LGTM. 
But @dubeejw would be best to vet the code :-)

dubeejw:
LGTM after the uuid change.

bjustin-ibm:
what change? 😝

PR #75, opened 2016-12-01
Add the test cases

houshengbo:


tests/src/system/packages/MessagingFeedTests.scala
```diff
@@ -0,0 +1,190 @@
...
+    it should "fire a trigger when a message is posted to the message hub" in withAssetCleaner(wskprops) {
+        // Load the credentials from the vcap file.
+        var credentials = TestUtils.getCredentials("message_hub")
+        val user = credentials.get("user").getAsString()
+        val password = credentials.get("password").getAsString()
+        val kafka_admin_url = credentials.get("kafka_admin_url").getAsString()
+        val api_key = credentials.get("api_key").getAsString()
+        val kafka_brokers_sasl_json_array = credentials.get("kafka_brokers_sasl").getAsJsonArray()
+
+        var vec = Vector[JsString]()
```
bjustin-ibm: This array juggling is pretty messy, but I assume it is unavoidable. Let's move this out somewhere that it can be reused. Perhaps a MessageHubCredentials class that does all this work in a `getKafkaBrokersSasl()` method that returns whatever type is directly needed for use in `trigger.create()`.

tests/src/system/packages/MessagingFeedTests.scala
```diff
@@ -0,0 +1,190 @@
...
+            producer.send(record)
+            producer.close()
+            
+            // Check if the trigger is fired.
+            val activations = wsk.activation.pollFor(N = 1, Some(triggerName), retries = 30).length
+            withClue("Change feed trigger count: ") { activations should be(1) }
+            
+            // Reset the property and delete and temporary configuration file.
+            System.setProperty("java.security.auth.login.config", "")
+            new File(tempConfigFile).delete()
```
bjustin-ibm: Since we will likely have more tests that interact with MessageHub, let's move this to a method that runs after all tests have completed.

tests/src/system/packages/MessagingFeedTests.scala
```diff
@@ -0,0 +1,190 @@
...
+
+            // Create a trigger with the Kafka feed.
+            var kafka_brokers = JsArray(JsString(config.kafkaHost))
+            val feedCreationResult = assetHelper.withCleaner(wsk.trigger, triggerName) {
+                (trigger, name) =>
+                    trigger.create(name, feed = Some(s"$messagingPackage/$kafkaFeed"), parameters = Map(
+                        "brokers" -> kafka_brokers,
+                        "topic" -> kafka_topic.toJson))
+            }
+            feedCreationResult.stdout should include("ok")
```
bjustin-ibm: While I'm sure you have seen this check done in other tests, it is not very robust. For example, if stdout contains "The result was **not** ok" then this check will pass because it only looks for the string "ok" somewhere in stdout.

I really like using `withActivation()` followed by `activation.response.success shouldBe true`. `withActivation` does the polling for you, as well as results in an object that is already parsed and ready for thorough inspection.

You can find a good example here: https://github.com/openwhisk/openwhisk/blob/master/tests/src/system/basic/Swift3WhiskObjectTests.scala#L56

tests/src/system/packages/MessagingFeedTests.scala
```diff
@@ -0,0 +1,190 @@
...
+            
+            // Give the consumer some to initialize.
+            Thread.sleep(2000)
+
+            // Send a message to the Kafka.
+            val message = new Message { override val serialize = s"${System.currentTimeMillis}" }
+            val sent = Await.result(producer.send(kafka_topic, message), 10 seconds)
+            
+            // Check if the trigger is fired.
+            val activations = wsk.activation.pollFor(N = 1, Some(triggerName), retries = 30).length
```
bjustin-ibm: See comment above about `withActivation()`

tests/src/system/packages/MessagingFeedTests.scala
```diff
@@ -0,0 +1,190 @@
...
+            
+            // Send a message to the message hub service.
+            val producer = new KafkaProducer[String, String](props)
+            val record = new ProducerRecord(topic, "key", currentTime)
+            producer.send(record)
+            producer.close()
+            
+            // Check if the trigger is fired.
+            val activations = wsk.activation.pollFor(N = 1, Some(triggerName), retries = 30).length
+            withClue("Change feed trigger count: ") { activations should be(1) }
```
bjustin-ibm: You should additionally check that the content of the activation contains the message you posted, as well as the other expected fields: topic, partition, offset, etc.

tests/src/system/packages/MessagingFeedTests.scala
bjustin-ibm: You should additionally check that the content of the activation contains the message you posted, as well as the other expected fields: topic, partition, offset, etc. You won't necessarily know what all the values will be, but you should at least check that those fields are present in the result.

tests/src/system/health/MessagingFeedTests.scala
```diff
@@ -0,0 +1,148 @@
...
+
+    val topic = "test"
+    val sessionTimeout = 10 seconds
+
+    implicit val wskprops = WskProps()
+    val wsk = new Wsk()
+
+    val messagingPackage = "/whisk.system/messaging"
+    val messageHubFeed = "messageHubFeed"
+
```
dubeejw: Add this method.
    
    def getNonDefaultNamespace(): String = {
        wsk.namespace.list().stdout.trim.split("\n").last
    }

tests/src/system/health/MessagingFeedTests.scala
```diff
@@ -0,0 +1,148 @@
...
+    	          new AppConfigurationEntry (
+    	              "com.ibm.messagehub.login.MessageHubLoginModule",
+     			          AppConfigurationEntry.LoginModuleControlFlag.REQUIRED, map))
+        })
+    }
+
+    behavior of "Message Hub"
+
+    it should "fire a trigger when a message is posted to the message hub" in withAssetCleaner(wskprops) {
+        var credentials = TestUtils.getCredentials("message_hub")
```
dubeejw: Add this value:
        val nondefaultNamespace = getNonDefaultNamespace()

tests/src/system/health/MessagingFeedTests.scala
```diff
@@ -0,0 +1,148 @@
...
+        System.setProperty("java.security.auth.login.config", "")
+        setMessageHubSecurityConfiguration(user, password)
+        var props = new Properties()
+        props.put("bootstrap.servers", servers);
+        props.put("security.protocol", "SASL_SSL");
+        props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
+        props.put("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
+
+        (wp, assetHelper) =>
+            val triggerName = s"dummyMessageHubTrigger-$currentTime"
```
dubeejw: Update triggerName:
            val triggerName = s"/$nondefaultNamespace/dummyMessageHubTrigger-$currentTime"

PR #78, opened 2016-12-02
Optionally disable SSL cert checking

### LINKED ISSUE #77: Disable SSL checking for local development
Often when developing/testing locally, OpenWhisk uses a self-signed certificate. This causes problems for this feed provider which (by default) insists on checking the validity of the SSL cert. With checking enabled during local development, GET and POST calls to OpenWhisk fail.

Fix will be to optionally disable SSL cert checking, controlled by the use of an environment variable.

bjustin-ibm:
Resolves #77

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -14,6 +14,7 @@
 
 import json
 import logging
+import os
 import requests
 import time
 
@@ -24,6 +25,8 @@ from database import Database
 from datetime import datetime
 from threading import Thread, Lock
 
+local_dev = os.getenv('LOCAL_DEV', 'False')
+check_ssl = (local_dev == 'False')
 
 class Consumer:
     class State:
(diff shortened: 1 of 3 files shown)


provider/consumer.py
```diff
@@ -24,6 +25,8 @@
 from datetime import datetime
 from threading import Thread, Lock
 
+local_dev = os.getenv('LOCAL_DEV', 'False')
```
dubeejw: Not sure that you need to check this variable twice. Just pass in the check_ssl variable during consumer creation?

bjustin-ibm: `LOCAL_DEV` is a variable that could be utilized by other parts of the full OpenWhisk deployment (whereas `check_ssl` not so much).  Additionally, `os.getenv()` returns the value as a String, which needs to be transformed into a Boolean anyway. Regardless of what the variable is called, we still need two lines of code to take the environment value and make it a Boolean.

PR #82, opened 2016-12-06
Do not Allow the default namespace

dubeejw:
- Check to ensure namespace is not _, an empty string, or undefined

Closes: https://github.com/openwhisk/openwhisk-package-kafka/issues/81

### DIFF
--- a/action/kafkaFeed.js
+++ b/action/kafkaFeed.js
@@ -17,6 +17,11 @@ function main(params) {
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
+    if (namespace === "_") {
+        whisk.error('You must supply a non-default namespace.');
+        return;
+    }
+
     var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
 
     if (params.lifecycleEvent === 'CREATE') {
--- a/action/messageHubFeed.js
+++ b/action/messageHubFeed.js
@@ -19,6 +19,11 @@ function main(params) {
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
+    if (namespace === "_") {
+        whisk.error('You must supply a non-default namespace.');
+        return;
+    }
+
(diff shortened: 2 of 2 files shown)


action/kafkaFeed.js
```diff
@@ -17,6 +17,11 @@ function main(params) {
     var namespace = encodeURIComponent(triggerComponents[1]);
     var trigger = encodeURIComponent(triggerComponents[2]);
 
+    if (namespace === '_') {
+        whisk.error('You must supply a non-default namespace.');
```
bjustin-ibm: @csantanapr What do you think about this error message? I was thinking maybe something like "Kafka triggers cannot be created in the default namespace. <tell them what do to instead>."

rabbah: the feed runs under the subject's namespace, so using the new properties is goodness. +1 @csantanapr's recommendation.

dubeejw:
Do we always want to use the environment to get the namespace, or only when the namespace is _?

csantanapr:
I would say to always use it.

action/kafkaFeed.js
dubeejw: Cannot use the environment variable as the user name be using a namespace with "_dev" in it, but process.env['__OW_NAMESPACE'] will not return a namespace with "_dev".

csantanapr:
@dubeejw I change my mind, it's correct to always use __OW_NAMESPACE 
The action will be ran with a specific namespace, that's that will be the string you will get.
If the run the action with `_` you will get the string for the default namespace which is not `_`

Sorry for all the flip flop

action/kafkaFeed.js
rabbah: This is not true with the new keys. One key = one valid org space.

dubeejw: Sure, but YP is still using the old keys.

dubeejw:
@csantanapr, the above is not the case when running against YP at least, which is what I'm testing against.

dubeejw:
If I create an action in the an _dev namespace, then create a trigger with a feed in my namespace that does not have _dev, the __OW_NAMESPACE environment variable will return the namespace containing _dev. That means a trigger will be created in my non-dev namespace, but the triggerURL variable generated in the feed action will point to a non-existent trigger in the dev namespace.

rabbah:
when we roll out the new keys this ambiguity goes away fwiw.

PR #88, opened 2016-12-09
Add the CouchDB support

houshengbo:
This PR adds CouchDB as a backend database support.

### DIFF
--- a/provider/database.py
+++ b/provider/database.py
@@ -17,15 +17,23 @@ import os
 import uuid
 
 from cloudant import Cloudant
+from cloudant import CouchDB
 from cloudant.result import Result
 
 class Database:
     db_prefix = os.getenv('DB_PREFIX', '')
     dbname = db_prefix + 'ow_kafka_triggers'
-    username = os.environ['CLOUDANT_USER']
-    password = os.environ['CLOUDANT_PASS']
-
-    client = Cloudant(username, password, account=username)
+    db_provider = os.environ.get('DB_PROVIDER', "cloudant")
+    username = os.environ.get('DB_USERNAME')
+    password = os.environ.get('DB_PASSWORD')
+    if ("cloudant" == db_provider.lower()):
+        client = Cloudant(username, password, account=username)
+    else:
+        db_host = os.environ.get('DB_HOST')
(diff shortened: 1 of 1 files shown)


provider/database.py
```diff
@@ -17,15 +17,30 @@
...
 class Database:
     db_prefix = os.getenv('DB_PREFIX', '')
     dbname = db_prefix + 'ow_kafka_triggers'
-    username = os.environ['CLOUDANT_USER']
-    password = os.environ['CLOUDANT_PASS']
+    username = os.environ.get('CLOUDANT_USER')
+    client = None
 
-    client = Cloudant(username, password, account=username)
+    if (username):
```
bjustin-ibm: Why not check the value of `DB_PROVIDER`? Above it seems odd to get the `CLOUDANT_USER` variable just to do this check. On first reading, I thought that `CLOUDANT_USER` was required. I understand that it is not required, but it made the code harder to understand.

provider/database.py
```diff
@@ -17,15 +17,30 @@
 import uuid
 
 from cloudant import Cloudant
+from cloudant import CouchDB
```
bjustin-ibm: Does the Dockerfile need to be updated to include installing this package?

provider/database.py
bjustin-ibm: Also, why keep `CLOUDANT_USER` _and_ `DB_USERNAME`? Seems like we only need one of these.

provider/database.py
houshengbo: No. It is the same library as cloudant.

provider/database.py
houshengbo: This is the tricky part, since bluewhisk is injecting CLOUDANT_USER and CLOUDANT_PASSWORD.
If we get rid of them, the install scripts in blue should be revised as well.

houshengbo:
@bjustin-ibm I have done the changes.

csantanapr:
@bjustin-ibm do we want to merge this? If so @houshengbo would need to resolve the conflict

bjustin-ibm:
@csantanapr No. CouchDB support will come along with #160

PR #89, opened 2016-12-09
Add Test Infrastructure

dubeejw:
- Add messageHubFeed tests
    - Add tests for service container
    - Demonstrate how to run test suite in the README


dubeejw:
@bjustin-ibm, I comment out the Kafka tests as I'm not sure if the Kafka feed will be installed in all environments.

bjustin-ibm:
@dubeejw I suspect the way to handle that is to exclude the generic kafka tests in those environments.

dubeejw:
@bjustin-ibm, please review.

dubeejw:
@bjustin-ibm, please try to run the tests again.

tests/src/system/packages/KafkaFeedTests.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  it should "reject invocation when topic argument is missing" in withAssetCleaner(wskprops) {
+    (wp, assetHelper) =>
+      val actionName = "missingTopicAction"
+      val expectedOutput = JsObject(
+        "error" -> JsString("You must supply a \"topic\" parameter.")
+      )
+
+      assetHelper.withCleaner(wsk.action, actionName) {
+        (action, _) => action.create(actionName, Some(actionFile))
```
bjustin-ibm: Rather than create (and delete) the action in every test, use the `BeforeAndAfterAll` trait to run a function before all the tests to create the action, and delete it after all the tests.

tests/src/system/packages/KafkaFeedTests.scala
```diff
@@ -0,0 +1,125 @@
...
+      }
+
+      val run = wsk.action.invoke(actionName, parameterFile = Some("dat/missingTopic.json"))
+
+      withActivation(wsk.activation, run) {
+        activation =>
+          activation.response.result shouldBe Some(expectedOutput)
```
bjustin-ibm: Please assert that the activation success is `false`.

bjustin-ibm: Also, the code for running the action and checking the result is repeated in every test. Let's create a function like `runActionWithExpectedResult(actionName, expectedResult, shouldSucceed=true)`

bjustin-ibm: You probably want to define that function in a separate trait/file so it can be reused in all your test classes.

tests/src/system/packages/MessagingServiceTests.scala
```diff
@@ -0,0 +1,218 @@
...
+    with Matchers {
+
+  val healthEndpoint = s"/health"
+
+  val getMessagingAddress =
+    if (System.getProperty("host") != "" && System.getProperty("port") != "") {
+      "http://" + System.getProperty("host") + ":" + System.getProperty("port")
+    }
+
+  behavior of "Provider endpoint"
```
bjustin-ibm: Please use a more specific name like "Messaging feed provider endpoint"

bjustin-ibm:
Rebased/squashed/merged as a2cfc972af76a2a32ab7e66170dae09599947bd8

PR #91, opened 2016-12-12
Make generic kafka functionality optional

bjustin-ibm:
Use an environment variable named `GENERIC_KAFKA` to control whether triggers can be created for unauthenticated Kafka instances. If the value is set to "True" generic Kafka triggers will be permitted. Any other values will result in a 403 response when attempting to create such triggers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ RUN pip install gevent==1.1.2 flask==0.11.1 confluent-kafka==0.9.2 \
 # set reasonable defaults
 ENV PORT 5000
 ENV LOCAL_DEV False
+ENV GENERIC_KAFKA True
 
 RUN mkdir -p /KafkaFeedProvider
 ADD provider/*.py /KafkaFeedProvider/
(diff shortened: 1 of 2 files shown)


provider/app.py
```diff
@@ -66,6 +66,14 @@ def postTrigger(namespace, trigger):
             'error': "trigger and namespace from route must correspond to triggerURL"
         })
         response.status_code = 409
+    elif body["isMessageHub"] == False and (enable_generic_kafka == False):
```
dubeejw: Why parentheses on the second condition?

Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN pip install gevent==1.1.2 flask==0.11.1 confluent-kafka==0.9.2 \
 # set reasonable defaults
 ENV PORT 5000
 ENV LOCAL_DEV False
+ENV GENERIC_KAFKA True
```
dubeejw: Do we plan on disabling generic Kafka in certain circumstances? If so, when?

provider/app.py
houshengbo: To be more pythonic, we should replace Line 69 with 
elif not body["isMessageHub"] and not enable_generic_kafka:

PR #95, opened 2016-12-13
Handle Trigger Hosts that do not Response

dubeejw:
- Catch error communicating with trigger host that occur while attempting to authenticate a trigger

Closes https://github.com/openwhisk/openwhisk-package-kafka/issues/94

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -71,16 +71,33 @@ def postTrigger(namespace, trigger):
         logging.warn("[{}] Attempt to create generic kafka trigger while function is disabled".format(triggerFQN))
         response = jsonify({
             'success': False,
-            'error': "Only triggers for Message Hub instances are allowed."
+            'error': "only triggers for Message Hub instances are allowed."
         })
         response.status_code = 403
     else:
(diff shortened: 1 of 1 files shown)


provider/app.py
```diff
@@ -71,16 +71,26 @@ def postTrigger(namespace, trigger):
...
         logging.info("[{}] Ensuring user has access rights to post a trigger".format(triggerFQN))
-        trigger_get_response = requests.get(body["triggerURL"], verify=check_ssl)
+
+        try:
+            trigger_get_response = requests.get(body["triggerURL"], verify=check_ssl)
+        except:
+            logging.warn("[{}] Failed to communicate with trigger host for authentication".format(triggerFQN))
+            response = jsonify({
+                'success': False,
+                'error': "failed to communicate with trigger host for authentication."
```
bjustin-ibm: Instead of "trigger host" let's refer it as "the OpenWhisk server". Maybe include the host and port in the message if it is easy to parse out of the `triggerURL` (https://docs.python.org/2/library/urlparse.html)

PR #97, opened 2016-12-14
Handle the case where params.endpoint already contains the protocol

### LINKED ISSUE #93: feed actions should tolerate an apihost value that already contains the protocol
In order to make the feed action deployment more flexible, the feed action should tolerate it when the apihost already contains the protocol (e.g. starts with "https://")

### LINKED ISSUE #96: refactor feed actions for code reuse
The two feed actions contain a lot of duplicate code. These should be refactored so that common code can go into a common file that is shared between the two actions. This will require deploying the feed actions as zip files.

bjustin-ibm:
If the protocol already exists, remove it so that the trigger URL can be properly constructed.

Resolves #93

### DIFF
--- a/action/kafkaFeed.js
+++ b/action/kafkaFeed.js
@@ -33,7 +33,10 @@ function main(params) {
         }
 
         var body = validatedParams;
-        body.triggerURL = 'https://' + whisk.getAuthKey() + "@" + params.endpoint + '/api/v1/namespaces/' + namespace + '/triggers/' + trigger;
+        // params.endpoint may already include the protocol - if so,
+        // strip it out
+        var massagedAPIHost = massageAPIHost(params.endpoint);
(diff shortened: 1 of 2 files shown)


action/kafkaFeed.js
```diff
@@ -120,3 +123,13 @@ function validateParameters(rawParams) {
 function isNonEmptyArray(obj) {
     return obj && Array.isArray(obj) && obj.length !== 0;
 }
+
+// if the apiHost already includes the protocol, remove it
+function massageAPIHost(apiHost) {
+    if(apiHost.substr(0, 4) === 'http') {
```
dubeejw: What about https? Could use regex to filter both http and https.

dubeejw: What if my hostname starts with `http`? Ex: httpmyhost.

bjustin-ibm: @dubeejw @rabbah Check it out: 4b98a5c11e39c17b5c12a682db4dc7376ac2e456

PR #110, opened 2017-01-06
remove use of deprecated whisk object

### LINKED ISSUE #98: whisk object deprecated
the actions in this package use the now deprecated whisk object - migration to promises (and npm openwhisk) should be in order.

bjustin-ibm:
There are two uses of of the (soon-to-be-removed) whisk object: `whisk.error()` and `whisk.getAuthKey()`

`whisk.error()` is replaced with promise rejection, and `whisk.getAuthKey()` is replaced with the `__OW_API_KEY` environment variable. Promise rejection required a little bit of refactoring to make it work nicely.

Resolves #98

### DIFF
--- a/action/kafkaFeed.js
+++ b/action/kafkaFeed.js
@@ -27,7 +27,7 @@ function main(params) {
                 // params.endpoint may already include the protocol - if so,
                 // strip it out
                 var massagedAPIHost = params.endpoint.replace(/https?:\/\/(.*)/, "$1");
-                body.triggerURL = 'https://' + whisk.getAuthKey() + "@" + massagedAPIHost + '/api/v1/namespaces/' + namespace + '/triggers/' + trigger;
+                body.triggerURL = 'https://' + process.env['__OW_API_KEY'] + "@" + massagedAPIHost + '/api/v1/namespaces/' + namespace + '/triggers/' + trigger;
 
                 var options = {
                     method: 'PUT',
@@ -47,7 +47,7 @@ function main(params) {
                 .then(resolve)
                 .catch(reject);
         } else if (params.lifecycleEvent === 'DELETE') {
-            var authorizationHeader = 'Basic ' + new Buffer(whisk.getAuthKey()).toString('base64');
+            var authorizationHeader = 'Basic ' + new Buffer(process.env['__OW_API_KEY']).toString('base64');
 
             var options = {
                 method: 'DELETE',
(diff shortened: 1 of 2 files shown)


action/messageHubFeed.js
```diff
@@ -52,25 +42,38 @@ function main(params) {
...
+            };
 
+            doRequest(options)
+                .then(resolve)
+                .catch(reject);
+        }
+    });
+
+    return promise;
+}
```
dubeejw: Need a new line after this bracket and before the next function.

action/kafkaFeed.js
```diff
@@ -8,60 +8,68 @@ var request = require('request');
...
-            }
-        };
+        var triggerComponents = params.triggerName.split("/");
+        var namespace = encodeURIComponent(process.env['__OW_NAMESPACE']);
+        var trigger = encodeURIComponent(triggerComponents[2]);
+        var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
+
+        if (params.lifecycleEvent === 'CREATE') {
+            // makes a PUT request to create the trigger in the feed service provider
+            var putTrigger = function(validatedParams) {
```
dubeejw: It may be cleaner to break this out into its own function outside of main().

action/messageHubFeed.js
```diff
@@ -10,36 +10,26 @@ var request = require('request');
...
-            .then(function() {
-                console.log("Successfully authenticated with Message Hub");
+        var triggerComponents = params.triggerName.split("/");
+        var namespace = encodeURIComponent(process.env['__OW_NAMESPACE']);
+        var trigger = encodeURIComponent(triggerComponents[2]);
+        var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
 
+        if (params.lifecycleEvent === 'CREATE') {
+            // makes a PUT request to create the trigger in the feed service provider
+            var putTrigger = function(validatedParams) {
```
dubeejw: It may be cleaner to break this out into its own function outside of main().

bjustin-ibm: I kind of agree. Inner functions are weird and confusing.

That being said, this function isn't used anywhere else _and_ it captures the `triggerURL` and `namespace` vars declared above.

dubeejw: You could pass those variables in as arguments?

PR #116, opened 2017-01-12
Prevent Race Condition on Consumer Delete

dubeejw:
- Use stopping state to prevent Doctor from causing a race condition when a consumer is stopped on purpose


dubeejw:
@bjustin-ibm, please review.

PR #117, opened 2017-01-13
Automatically restart the container if the health endpoint fails

### LINKED ISSUE #115: Use Docker Healthcheck to automatically restart the container if needed
https://docs.docker.com/engine/reference/builder/#/healthcheck

The `healthcheck` instruction tells Docker how to determine if the container is unhealthy, even if it is still running. For us, we should use this to `curl` the `health` endpoint to make sure the feed provider is still responsive and automatically restart it if it does not respond in a reasonable amount of time.

bjustin-ibm:
Use the docker HEALTHCHECK instruction to curl the /health endpoint every five minutes.
If the endpoint does not respond within 30 seconds, cause the container's restart policy to kick in by killing the main python process.
Unfortunately, docker 1.12 does not (yet) support automatically restarting containers in response to healthcheck failures. This is a stop-gap measure to get the container to restart.

Closes #115

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ RUN apt-get install -y \
     make \
     zlib1g-dev \
     libsasl2-dev \
-    libsasl2-modules
+    libsasl2-modules \
+    psmisc
 
 # install librdkafka
 RUN git clone --depth 1 --branch v0.9.2 https://github.com/edenhill/librdkafka.git librdkafka \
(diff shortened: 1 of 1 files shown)


bjustin-ibm:
@dubeejw Please review and merge.

dubeejw:
@bjustin-ibm, is there a way to alert us that the provider has been restarted, so that we can investigate why the restart happened?

bjustin-ibm:
This was rebased/merged as 31af4fbee6369d2eb113761457c215505cab2293

Issue #118, opened 2017-01-15
logged message does not contain partition and offset


bjustin-ibm:
When messages are intentionally skipped due to a failed number of retries firing the trigger (usually due to a throttled response) the logged message is intended to contain the partition and offset that will be skipped. However, due to a coding error, this message is incorrectly formatted:

```
Skipping 1 messages to offset <built-in method offset of cimpl.Message object at 0x7fd8a420b470> of partition <built-in method partition of cimpl.Message object at 0x7fd8a420b470>
```

The problem is that `offset` and `partition` are functions and not fields.

PR #120, opened 2017-01-16
Add additional INFO logging

bjustin-ibm:
Log when consumer completes their first poll()
Log when consumer is closed and deallocated


bjustin-ibm:
I'm gonna roll these into #118

PR #121, opened 2017-01-16
log the activation when successfully firing a trigger

bjustin-ibm:


provider/consumer.py
```diff
@@ -313,6 +313,7 @@ def __fireTrigger(self, messages):
                     # service was unavailable (503), an internal server error occurred (500), the request timed out
                     # (408), or the request was throttled (429).
                     if status_code == 200:
+                        logging.info("[{}] Fired trigger with activation {}".format(self.trigger, response.json()['activationId']))
```
dubeejw: Should we also log the activation when the trigger was not successfully fired and an activationId exists in the response?

bjustin-ibm: Is it possible to get a non-200 response with an activation id?

PR #127, opened 2017-01-31
Actions to produce messages to Kafka and Message Hub

### LINKED ISSUE #18: action to post a message to Kafka
The Kafka/Message Hub package should include an action that allows the caller to post a message to some topic in some Kafka or Message Hub instance.

bjustin-ibm:
Resolves #18

### DIFF
--- a/action/KafkaProduce/postKafka.js
+++ b/action/KafkaProduce/postKafka.js
@@ -1,0 +1,38 @@
+/**
+ *  Action to post a message to a Message Hub instance
+ *  @param {string} brokers - array of Kafka brokers
+ *  @param {string} topic - topic to produce message
+ *  @param {string} value - content of message
+ *  @param {string} key - message key (optional)
+ *  @param {int}    required_acks - number of acks required before action is complete (optional, default = 1)
+ *  @param {bool}   isJSONData - attempt to parse messages as JSON
+ */
+function main(args) {
+    return validateParameters(args)
+        .then(produceMessage);
+}
+
+function produceMessage(args) {
+    const produce = require('../lib/produce');
+
+    var producerConfig = {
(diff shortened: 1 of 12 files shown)


bjustin-ibm:
@dubeejw This is not 100% ready yet, but I would appreciate your input so far.

dubeejw:
Made some comments, still going through it though.

action/KafkaProduce/postKafka.js
```diff
@@ -0,0 +1,38 @@
...
+ *  @param {string} brokers - array of Kafka brokers
+ *  @param {string} topic - topic to produce message
+ *  @param {string} value - content of message
+ *  @param {string} key - message key (optional)
+ *  @param {int}    required_acks - number of acks required before action is complete (optional, default = 1)
+ *  @param {bool}   isJSONData - attempt to parse messages as JSON
+ */
+function main(args) {
+    return validateParameters(args)
+        .then(produceMessage);
```
bjustin-ibm: I think you are referring to `produceMessage` not being called with any arguments. If so, notice how it is not being called at all, but instead the function itself is being passed as an argument to `then()`. The result of this is that when the `validateParameters` promise is resolved, `produceMessage` will be called with whatever the `validateParameters` promise resolved to as an argument (in this case, the arguments that were validated by `validateParameters`).

For example:
```javascript
function main() {
    return doTheThing()
        .then(sayTheThing);
}

function doTheThing() {
    return Promise.resolve("Hey!");
}

function sayTheThing(message) {
    console.log("You told me to say: " + message);
}
```

Invoking `main()` eventually prints the message `You told me to say: Hey!`

On the other hand, I might be totally misunderstanding your comment. Let me know...

bjustin-ibm:
Superseded by #131

PR #130, opened 2017-02-06
Fix tests

bjustin-ibm:
Consolidate all MH initialization into a new KafkaUtils class.

Check the resulting activations to make sure that one of them actually contains the keys and values for the message produced by the test. This protects against running the test concurrently on the same MH queue.


bjustin-ibm:
Superseded by #140

Issue #125, opened 2017-01-22
support for binary message payload


glikson:
Currently, the message value must be a valid utf-8 string (passed in the JSON payload of the trigger). Removing this limitation would enable many important use-cases for the Kafka feed.
For example, isBinaryData flag could be added, in which case the message value could be base64-encoded.

dubeejw:
@glikson, I am able to reproduce your issue by explicitly encoding the message to be sent from the producer as utf-8. If I don't encode the message as utf-8, I do not receive any error. See the example below. 

Can you show us your producer code?

Produces error:
```
message = bytes([0x8b])
producer.send(topic, message.encode('utf-8'))
```

Does not produce error:
```
message = bytes([0x8b])
producer.send(topic, message)
```

glikson:
It is not my producer - it is coming from the bridge from IoT Platform to Message Hub. The messages are just binary, not utf-8 compliant. So, an attempt to parse them as utf-8 (which was the assumption of MH feed code until now, as far as I understand) was very likely to yield an exception.

dubeejw:
@glikson, the exception is actually thrown from `response = requests.post(self.triggerURL, json=payload, timeout=10.0, verify=check_ssl)`. I'll encode the messages in Base64 before reaching this method call.

dubeejw:
@glikson, from the above post, I meant that I can raise an exception in the consumer by encoding the message in utf-8 from the producer.

bjustin-ibm:
Binary data support for the produce actions is handled by #140

PR #132, opened 2017-02-09
Support Base64 Encoding of Binary Message Payloads

dubeejw:
- Add decoding for binary messages encode in base64

Closes https://github.com/openwhisk/openwhisk-package-kafka/issues/125


dubeejw:
@bjustin-ibm, subclassing the Consumer class is not currently possible. There was a update to the Confluent Kafka library recently to allow subclassing of the Consumer and Producer. However, that change is not in a released version of the library, it is only in master. See: https://github.com/confluentinc/confluent-kafka-python/commit/864a9358a6464a1a8aa50743859fe13e9d85381d

bjustin-ibm:
@dubeejw As nice as this test is, I would really like to try it with a message value that we know will cause an exception without this code. I am a little skeptical that this current approach actually solves that issue.

According to @glikson, he found the problem using a gzipped JSON array.

dubeejw:
@bjustin-ibm, was an exception the result, or just malformed user input?

bjustin-ibm:
```
Uncaught exception: 'utf8' codec can't decode byte 0x8b in position 1: invalid start byte
```

IIRC that exception was thrown from `poll()` but I can't be sure. My gut feeling is that the exception is thrown from inside the Python kafka consumer.

README.md
```diff
@@ -33,7 +34,8 @@ In order to create a trigger that reacts when messages are posted to an unauthen
 |---|---|---|
 |brokers|JSON Array of Strings|This parameter is an array of `<host>:<port>` strings which comprise the brokers in your Message Hub instance|
 |topic|String|The topic you would like the trigger to listen to|
-|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the feed to attempt the message content as JSON before passing it along as the trigger payload.|
+|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to parse the message content as JSON before passing it along as the trigger payload.|
+|isBase64Data|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to decode the message content as Base64 before passing it along as the trigger payload.|
```
glikson: As useful as this might be, it is the opposite of the issue I have been reporting in #125. The flag should cause the feed to *encode* incoming binary data as base64, so that it can be passed as part of a valid JSON to the action.

provider/consumer.py
```diff
@@ -349,13 +354,19 @@ def __parseMessageIfNeeded(self, value):
...
                 return parsed
             except ValueError:
                 # no big deal, just return the original value
-                logging.warn(
-                    '[{}] I was asked to parse a message as JSON, but I failed.'.format(self.trigger))
+                logging.warn('[{}] I was asked to parse a message as JSON, but I failed.'.format(self.trigger))
+                pass
+        elif self.parseAsBase64:
+            try:
+                parsed = value.decode("base64")
```
glikson: ditto - should be 'encode' into base64 rather than 'decode' from base64.

glikson:
See my comments in the code -- this is not solving the problem I am facing. The format transformation should happen in the opposite direction -- from binary (not utf-8 compliant, so can't be passed in the trigger payload JSON) to base64 (which *is* utf-8 compliant, and the conversion is lossless), rather than from base64 to utf-8.

glikson:
I believe the error happens as a result of this line in the consumer:
'value': self.__parseMessageIfNeeded(message.value())

it assumes that the right-hand side is a JSON string (utf-8 compliant), while causes an exception if the message is binary.

dubeejw:
@bjustin-ibm, ready for review.

README.md
```diff
@@ -88,6 +92,39 @@ However, if the same message content is posted with `isJSONData` set to `false`,
...
+            "key": null,
+            "offset": 421760,
+            "topic": "mytopic",
+            "value": "U29tZSBkYXRh"
+        }
+    ]
+}
+```
+
+If the same message is posted without `isBinaryData` set to `true`, the trigger payload would resemble the below example:
```
bjustin-ibm: This counter example is probably not necessary.

provider/consumer.py
```diff
@@ -349,13 +354,19 @@ def __parseMessageIfNeeded(self, value):
...
                 return parsed
             except ValueError:
                 # no big deal, just return the original value
-                logging.warn(
-                    '[{}] I was asked to parse a message as JSON, but I failed.'.format(self.trigger))
+                logging.warn('[{}] I was asked to parse a message as JSON, but I failed.'.format(self.trigger))
+                pass
+        elif self.parseAsBinary:
+            try:
+                parsed = value.encode("base64").strip()
```
bjustin-ibm: Why do you need `strip()` here?

README.md
```diff
@@ -88,6 +92,39 @@ However, if the same message content is posted with `isJSONData` set to `false`,
 }
 ```
 
+Similar to `isJSONData`, if `isBinaryData` was set to `true` during trigger creation, an attempt will be made to encode values from binary to Base64. If successfully encoded, the resultant `value` in the trigger payload will be encoded as a Base64 string.
+
+For example, if a message of `Some data` is posted with `isBinaryData` set to `true`, the trigger payload might look something like this:
```
glikson: Not sure this example is accurate - "some data" is not binary.. maybe give an example with buffer.

README.md
```diff
@@ -15,7 +15,8 @@ In order to create a trigger that reacts when messages are posted to a Message H
 |password|String|Your Message Hub password|
 |topic|String|The topic you would like the trigger to listen to|
 |kafka_admin_url|URL String|The URL of the Message Hub admin REST interface|
-|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the feed to attempt the message content as JSON before passing it along as the trigger payload.|
+|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to parse the message content as JSON before passing it along as the trigger payload.|
+|isBinaryData|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to encode the message content as Base64 before passing it along as the trigger payload.|
```
glikson: Strictly speaking, I am not sure base64 encoding can ever fail.. if this is indeed the case, maybe drop the 'attempt' part..

provider/consumer.py
```diff
@@ -131,6 +131,11 @@ def __init__(self, trigger, params):
         else:
             self.parseAsJson = False
 
+        if "isBinaryData" in params:
+            self.parseAsBinary = params["isBinaryData"]
```
glikson: this is a nit, but "encoding" is not exactly the same as "parsing"..

provider/consumer.py
dubeejw: The encode function used here may add whitespace.

dubeejw:
@bjustin-ibm, `isBinaryKey` and `isBinaryValue` added.

README.md
```diff
@@ -15,7 +15,9 @@ In order to create a trigger that reacts when messages are posted to a Message H
 |password|String|Your Message Hub password|
 |topic|String|The topic you would like the trigger to listen to|
 |kafka_admin_url|URL String|The URL of the Message Hub admin REST interface|
-|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the feed to attempt the message content as JSON before passing it along as the trigger payload.|
+|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to parse the message content as JSON before passing it along as the trigger payload.|
+|isBinaryKey|Boolean (Optional - default=false)|When set to `true` this will cause the provider to encode the key content as Base64 before passing it along as the trigger payload.|
+|isBinaryValue|Boolean (Optional - default=false)|When set to `true` this will cause the provider to encode the message content as Base64 before passing it along as the trigger payload.|
```
bjustin-ibm: ~content~ value

README.md
```diff
@@ -33,7 +35,9 @@ In order to create a trigger that reacts when messages are posted to an unauthen
 |---|---|---|
 |brokers|JSON Array of Strings|This parameter is an array of `<host>:<port>` strings which comprise the brokers in your Message Hub instance|
 |topic|String|The topic you would like the trigger to listen to|
-|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the feed to attempt the message content as JSON before passing it along as the trigger payload.|
+|isJSONData|Boolean (Optional - default=false)|When set to `true` this will cause the provider to attempt to parse the message content as JSON before passing it along as the trigger payload.|
+|isBinaryKey|Boolean (Optional - default=false)|When set to `true` this will cause the provider to encode the key content as Base64 before passing it along as the trigger payload.|
+|isBinaryValue|Boolean (Optional - default=false)|When set to `true` this will cause the provider to encode the message content as Base64 before passing it along as the trigger payload.|
```
bjustin-ibm: ~content~ value

action/kafkaFeed.js
```diff
@@ -104,6 +106,13 @@ function validateParameters(rawParams) {
 
         validatedParams.isMessageHub = false;
         validatedParams.isJSONData = (typeof rawParams.isJSONData !== 'undefined' && rawParams.isJSONData && (rawParams.isJSONData === true || rawParams.isJSONData.toString().trim().toLowerCase() === 'true'));
+        validatedParams.isBinaryValue = (typeof rawParams.isBinaryValue !== 'undefined' && rawParams.isBinaryValue && (rawParams.isBinaryValue === true || rawParams.isBinaryValue.toString().trim().toLowerCase() === 'true'));
+        validatedParams.isBinaryKey = (typeof rawParams.isBinaryKey !== 'undefined' && rawParams.isBinaryKey && (rawParams.isBinaryKey === true || rawParams.isBinaryKey.toString().trim().toLowerCase() === 'true'));
+
+        if (validatedParams.isJSONData && validatedParams.isBinaryValue) {
```
bjustin-ibm: Seems like this ought to be in the `validateParameters` function

action/messageHubFeed.js
```diff
@@ -142,6 +144,13 @@ function validateParameters(rawParams) {
 
         validatedParams.isMessageHub = true;
         validatedParams.isJSONData = (typeof rawParams.isJSONData !== 'undefined' && rawParams.isJSONData && (rawParams.isJSONData === true || rawParams.isJSONData.toString().trim().toLowerCase() === 'true'));
+        validatedParams.isBinaryValue = (typeof rawParams.isBinaryValue !== 'undefined' && rawParams.isBinaryValue && (rawParams.isBinaryValue === true || rawParams.isJSONData.toString().trim().toLowerCase() === 'true'));
+        validatedParams.isBinaryKey = (typeof rawParams.isBinaryKey !== 'undefined' && rawParams.isBinaryKey && (rawParams.isBinaryKey === true || rawParams.isBinaryKey.toString().trim().toLowerCase() === 'true'));
+
+        if (validatedParams.isJSONData && validatedParams.isBinaryValue) {
```
bjustin-ibm: Same, should be in `validateParameters` function, no?

installCatalog.sh
```diff
@@ -45,7 +45,7 @@ echo Installing the kafka package and feed action.
...
+    -a parameters '[ {"name":"kafka_brokers_sasl", "required":true, "description": "Array of Message Hub brokers", "bindTime":true},{"name":"user", "required":true, "description": "Message Hub username", "bindTime":true},{"name":"password", "required":true, "description": "Message Hub password", "bindTime":true, "type":"password"},{"name":"topic", "required":true, "description": "Topic to subscribe to"},{"name":"isJSONData", "required":false, "description": "Attempt to parse message content as JSON"},{"name":"isBinaryKey", "required":false, "description": "Encode key content as Base64"},{"name":"isBinaryValue", "required":false, "description": "Encode message content as Base64"},{"name":"endpoint", "required":true, "description": "Hostname and port of OpenWhisk deployment"},{"name":"kafka_admin_url", "required":true, "description": "Your Message Hub admin REST URL", "bindTime":true}]' \
```
bjustin-ibm: hmmm... seems weird to document all these parameters on the package when they don't apply to all the actions in the package.

installCatalog.sh
```diff
@@ -54,7 +54,7 @@ $WSK_CLI -i --apihost "$EDGEHOST" action update messaging/messageHubFeed "$PACKA
...
+    -a parameters '[ {"name":"kafka_brokers_sasl", "required":true, "description": "Array of Message Hub brokers"},{"name":"user", "required":true, "description": "Message Hub username"},{"name":"password", "required":true, "description": "Message Hub password", "type":"password"},{"name":"topic", "required":true, "description": "Topic to subscribe to"},{"name":"isJSONData", "required":false, "description": "Attempt to parse message content as JSON"},{"name":"isBinaryKey", "required":false, "description": "Encode key content as Base64"},{"name":"isBinaryValue", "required":false, "description": "Encode message content as Base64"},{"name":"endpoint", "required":true, "description": "Hostname and port of OpenWhisk deployment"},{"name":"kafka_admin_url", "required":true, "description": "Your Message Hub admin REST URL"}]' \
```
bjustin-ibm: ~content~ value

provider/consumer.py
```diff
@@ -127,9 +127,19 @@ def __init__(self, trigger, params):
         # handle the case where there may be existing triggers that do not
         # have the isJSONData field set
         if "isJSONData" in params:
-            self.parseAsJson = params["isJSONData"]
+            self.encodeValueAsJSON = params["isJSONData"]
```
bjustin-ibm: heh... opposite nit: here we are parsing and not encoding 😝

dubeejw: According to the Python doc for the JSON library, json.loads() technically encodes JSON strings into a Python Object. See https://docs.python.org/2/library/json.html.

PR #134, opened 2017-02-14
Remove dead code from KafkaUtils

jberstler:
No longer need this to create a producer now that the health test uses the messageHubProduce action
This also means we can get rid of all the code to configure the JVM for use with Message Hub


bjustin-ibm:
Let's put this on the back burner for now as we may need the producer code in this class in order to test #125

PR #136, opened 2017-02-14
Added an image and updated docs in README.md

romankhar:
cc: @csantanapr


README.md
```diff
@@ -127,6 +166,18 @@ Here is an example of invoking the action with all required parameters:
 wsk action invoke /messaging/kafkaProduce -p brokers "[\"mykafkahost:9092\", \"mykafkahost:9093\"]" -p topic mytopic -p value "This is the content of my message"
 ```
 
+### Producing Messages to a Message Hub or Generic Kafka Instance using DIY action
```
bjustin-ibm: I'm not sure I understand the purpose of this section. It kind of seems unrelated to the messaging package in that it says "you really don't need this now, but just in case, here is an example of how you _used to_ do things."

It is a nice example though, so I wonder if maybe we just turn this into a more generic "Examples" section with a very brief (one sentence) description of the sample with a link to the GitHub project. If that becomes the case, then the previous two section titles can go back to the way they were before (without the "using system provided action").

romankhar: agreed - I made changes and pushed an update

csantanapr:
merged b24b88b4c00d2ea76f569d7589694026325fce9e

PR #140, opened 2017-02-16
produce actions should handle binary keys and values

bjustin-ibm:
Optionally base64 decode the message key and/or value through the use of two new action parameters.


bjustin-ibm:
@dubeejw When merging, please use the "rebase and merge" option (or do it manually) as I would like to keep these 2 commits separate. Thanks.

bjustin-ibm:
@glikson Feel free to comb through this one if you like.

action/kafkaProduce.py
```diff
@@ -16,11 +17,19 @@ def main(params):
 
         print "Created producer"
 
+        messageValue = params['value']
+        if 'base64DecodeValue' in params and params['base64DecodeValue'] == True:
+            messageValue = base64.b64decode(params['value'])
```
glikson: can this fail? what happens then?

tests/src/system/packages/MessageHubProduceTests.scala
```diff
@@ -116,4 +120,112 @@ class MessageHubProduceTests
...
+            }
+
+            withActivation(wsk.activation, feedCreationResult, initialWait = 5 seconds, totalWait = 60 seconds) {
+                activation =>
+                    // should be successful
+                    activation.response.success shouldBe true
+            }
+
+            // produce message
+            val decodedMessage = "This will be base64 encoded"
```
glikson: maybe consider using a buffer with actually binary data..

action/kafkaProduce.py
bjustin-ibm: Good one! This can fail and, as written when you made this comment, the invocation would have failed with potentially cryptic message. I have enhanced this situation to fail in a more informative way:

```json
    "response": {
        "status": "application error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "value parameter is not base64 encoded"
        }
    }
```

bjustin-ibm:
@dubeejw ready for review

action/kafkaProduce.py
```diff
@@ -16,11 +17,19 @@ def main(params):
...
+
         # only use the key parameter if it is present
         if 'key' in params:
-            producer.send(params['topic'], bytes(params['value']), key=bytes(params['key']))
+            messageKey = params['key']
+            if 'base64DecodeKey' in params and params['base64DecodeKey'] == True:
+                messageKey = base64.b64decode(params['key'])
```
dubeejw: Can just do `params['key'].decode('base64').strip()` instead of importing another library to do the decoding.

action/kafkaProduce.py
```diff
@@ -45,5 +54,19 @@ def validateParams(params):
...
+    elif 'base64DecodeKey' in params and params['base64DecodeKey'] == True and not isBase64(params['key']):
+        return (False, "key parameter is not base64 encoded")
     else:
         return (True, "Params all check out.")
+
+def isBase64(s):
+    print "In isBase64"
+    valid = True
+    try:
+        valid = (base64.b64encode(base64.b64decode(s)) == s)
```
dubeejw: If you use the `.decode("base64")` method, decoding values that result from trying to decode a non-base64 value are an empty string. It seems like it'd be must each to check string length instead of what you are doing here.

action/kafkaProduce.py
dubeejw: Seems like the decoding should be done before `validateParams()` is called, and the decoded result put into the params sent to the method. This way the decoding doesn't occur twice.

action/messageHubProduce.py
```diff
@@ -27,11 +28,19 @@ def main(params):
 
         print "Created producer"
 
+        messageValue = params['value']
+        if 'base64DecodeValue' in params and params['base64DecodeValue'] == True:
+            messageValue = base64.b64decode(params['value'])
```
dubeejw: Same comments form kafkaProduce.py apply to messageHubProduce.py.

tests/src/system/packages/MessageHubFeedTests.scala
```diff
@@ -113,4 +134,71 @@ class MessageHubFeedTests
     runActionWithExpectedResult(actionName, "dat/multipleValueTypes.json", expectedOutput, false)
   }
 
+  it should "fire a trigger when a binary message is posted to message hub" in withAssetCleaner(wskprops) {
```
dubeejw: Would be nice if we had a generic way to create a trigger, produce a value, and check that value from the fired trigger. Such a mechanism would easily allow us to run this end-to-end type test with minimal duplicate code as it is now in three different files.

tests/src/system/packages/MessageHubProduceTests.scala
```diff
@@ -116,4 +120,132 @@ class MessageHubProduceTests
...
+
+            val activation = matchingActivations.head
+            activation.getFieldPath("response", "success") shouldBe Some(true.toJson)
+
+            // assert that there exists a message in the activation which has the expected keys and values
+            val messages = KafkaUtils.messagesInActivation(activation, field = "value", value = decodedMessage)
+            assert(messages.length == 1)
+    }
+
+    it should "Post a message with a binary key" in withAssetCleaner(wskprops) {
```
dubeejw: Sure, producing an encoded key and value work individually, but do they work when both enabled?

action/kafkaProduce.py
dubeejw: Ex:
```
>>> invalid = "????????????".decode("base64")
>>> print "'%s'" % (invalid)
''
```

README.md
```diff
@@ -200,6 +200,8 @@ If you would like to use an OpenWhisk action to conveniently produce a message t
 |topic|String|The topic you would like the trigger to listen to|
 |value|String|The value for the message you would like to produce|
 |key|String (Optional)|The key for the message you would like to produce|
+|base64DecodeValue|Boolean (Optional - default=false)|If true, the message will be produced with a base64 decoded version of the value parameter|
```
dubeejw: Should probably capitalize all the occurrences of `base64` in this file to match the existing doc style.

README.md
```diff
@@ -254,21 +256,28 @@ If you would like to use an OpenWhisk action to conveniently produce a message t
...
+Before the time when OpenWhisk had built-in action to publish messages to and from Message Hub and Kafka developers used to accomplish this functionality using custom built actions using any of the OpenWhisk supported programming languages. This is still possible today and if, for some reason you are not satisfied with OpenWhisk provided Message Hub and Kafka feed and publish action, you can still develop and use your own. While creating such action, you can use Message Hub and Kafka Native API or Confluence REST API for Kafka. It is transparent to OpenWhisk which you use, just like with any other 3rd party API or library you decide to utilize in your OpenWhisk action. This [sample project](https://github.com/IBM/ogs-data-processing-messagehub) demonstrates the use of OpenWhisk for sending and receiving IBM Message Hub and Kafka messages.
```
dubeejw: Missing comma. Before the time when OpenWhisk had built-in action to publish messages to and from Message Hub and Kafka`,` developers

bjustin-ibm: bah! I didn't even change this 😛

bjustin-ibm: Actually this whole section is made obsolete by the new produce actions. I'm just going to remove it.

csantanapr: As reference this is what I'm currently pulling out for the bluemix docs, which is less what the readme contains.
https://console.ng.bluemix.net/docs/openwhisk/openwhisk_messagehub.html

Since kafka is not installed by default and not available in bluemix, maybe we move it to a secondary md file.

The closer the readme is to the bluemix docs and openwhisk.org (future) the better.

PR #143, opened 2017-03-02
upgrade to kafka client libs v0.9.4

### LINKED ISSUE #111: service failed to clean up consumer for deleted trigger
I observed an incident where the service became entirely unresponsive (HTTP requests not handled, messages not consumed, triggers not fired). The last meaningful log messages looked like this:

```
2017-01-08 02:16:45,225 [INFO]: [<TRIGGER>] Cleaning up consumer
%3|1483841836.386|FAIL|rdkafka#consumer-339| sasl_ssl:///<BROKER>/1: 
Receive failed: No error
%3|1483841836.386|ERROR|rdkafka#consumer-339| sasl_ssl:///<BROKER>/1:
 Receive failed: No error
%3|1483841862.127|FAIL|rdkafka#consumer-1| sasl_ssl://<BROKER>/0: Re
ceive failed: No error
```

Incidentally, `1483841836.386` is approximately `2017-01-08 02:16` so these messages are basically consecutive.

The messages from `rdkafka` repeat indefinitely and the expected log message `Successfully closed KafkaConsumer` is never seen.

bjustin-ibm:
Resolves #111

### DIFF
--- a/tests/src/system/stress/StressTest.scala
+++ b/tests/src/system/stress/StressTest.scala
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package system.stress
+
+import system.utils.KafkaUtils
+
+import scala.concurrent.duration.DurationInt
+import scala.language.postfixOps
+
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.Matchers
+import org.scalatest.junit.JUnitRunner
+
+import common.TestHelpers
+import common.Wsk
+import common.WskActorSystem
+import common.WskProps
+import common.WskTestHelpers
+import spray.json.DefaultJsonProtocol._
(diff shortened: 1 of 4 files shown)


tests/src/system/stress/StressTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+            // optionally allow triggers to pile up on the provider
+            if((iterationLabel % keepNthTrigger) != 0) {
+                println("Deleting trigger")
+                val feedDeletionResult = wsk.trigger.delete(triggerName)
+                feedDeletionResult.stdout should include("ok")
+                stressTriggerCreateAndDelete(totalIterations, keepNthTrigger, currentIteration + 1, storedTriggers)
+            } else {
+                println("I think I'll keep this trigger...")
+                stressTriggerCreateAndDelete(totalIterations, keepNthTrigger, currentIteration + 1, triggerName :: storedTriggers)
```
dubeejw: Could make this call once after the if statement instead of in each block of the if statement, and just append the new stored trigger name in the else.

bjustin-ibm: That would require a mutable list which is a no-no in Scala :)

tests/src/system/stress/StressTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+    /*
+     * Recursively create and delete (potentially) lots of triggers
+     *
+     * @param totalIterations The total number of triggers to create
+     * @param keepNthTrigger Optionally, do not delete the trigger created on every N iterations
+     * @param currentIteration Used for recursion
+     * @param storedTriggers The list of trigger names that were created, but not deleted (see keepNthTrigger)
+     */
+    def stressTriggerCreateAndDelete(totalIterations : Int, keepNthTrigger : Int, currentIteration : Int = 0, storedTriggers : List[String] = List[String]()) {
```
dubeejw: This test would be nice if it could create triggers in parallel.

bjustin-ibm: That may be true, but it isn't required to reproduce the bug in #111

PR #146, opened 2017-03-07
Use blocking invokes when running messageHubProduce

bjustin-ibm:
These tests really should not proceed until it is known that the produce action fully completed.

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -96,7 +96,7 @@ class BasicHealthTest
                 "kafka_brokers_sasl" -> kafkaUtils.getAsJson("brokers"),
                 "topic" -> topic.toJson,
                 "key" -> key.toJson,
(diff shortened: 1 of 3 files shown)


bjustin-ibm:
FYI @jasonpet this could very well explain the intermittent failures.

dubeejw:
Using blocking actions does not guarantee that the code will not proceed because the blocking action may exceed an execution time of 60 seconds and then turn into a non-blocking action.

bjustin-ibm:
Yes, I'm aware. Are you suggesting it is better to leave these invokes non-blocking? :trollface:

dubeejw:
The test, in its current configuration, will fail if the blocking invocation goes async.

markusthoemmes:
`withActivation` blocks until the it found the activation record as well, no?

bjustin-ibm:
> The test, in its current configuration, will fail if the blocking invocation goes async. The code will not move beyond the invocation call.

This seems reasonable to me. If the produce action takes longer than 60s then something is wrong and the test cannot proceed.

PR #149, opened 2017-03-13
Updating all Python files to have ASF licenses and pass PEP8 tests

mrutkows:
Required for Apache and to pass new scanCode.py tests going forward.

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -1,30 +1,32 @@
-# Copyright 2016 IBM Corp. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-# https://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+"""Flask application.
 
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+"""
 import logging
 import os
 import requests
-import sys
 import urllib
 import uuid
 
-from datetime import datetime
 from flask import Flask, jsonify, request
 from consumer import Consumer
 from consumercollection import ConsumerCollection
 from database import Database
-from threading import Lock
 from thedoctor import TheDoctor
 from urlparse import urlparse
 from health import generateHealthReport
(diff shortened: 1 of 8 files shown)


provider/app.py
```diff
@@ -154,15 +173,17 @@ def healthRoute():
...
 
 
 def createAndRunConsumer(triggerFQN, params, record=True):
-    if app.config['TESTING'] == True:
+    if app.config['TESTING'] is True:
         logging.debug("Just testing")
     else:
         # generate a random uuid for new triggers
-        if not 'uuid' in params:
+        if 'uuid' not in params:
```
bjustin-ibm: Is this change necessary? This change is grammatically less readable.

provider/app.py
```diff
@@ -205,7 +226,8 @@ def main():
 
     # Make sure we log to the console
     streamHandler = logging.StreamHandler()
-    formatter = logging.Formatter('[%(asctime)s] [%(levelname)s] [??] [kafkatriggers] %(message)s')
+    formatter = logging.Formatter('[%(asctime)s] [%(levelname)s] '
```
bjustin-ibm: Are all these formatting-only changes necessary? If not, I would suggest removing them since they are just muddying the history of this file. If they are necessary, then what is to prevent us from creating more formatting issues in the future?

provider/consumer.py
```diff
@@ -65,20 +72,25 @@ def start(self):
...
         with self.__lock:
             self.__restartCount += 1
 
-        logging.info('[{}] Quietly shutting down consumer for restart'.format(self.trigger))
+        logging.info("[{}] Quietly shutting down consumer for "
+                     "restart".format(self.trigger))
         self.thread.setDesiredState(Consumer.State.Restart)
         self.thread.join()
-        logging.info('Consumer has shut down')
+        logging.info("Consumer has shut down")
```
bjustin-ibm: hmmm... quote usage throughout these changes is inconsistent. What is the reason for changing some but not others?

provider/consumer.py
```diff
@@ -201,58 +219,71 @@ def run(self):
                 if len(messages) > 0:
                     self.__fireTrigger(messages)
 
-            logging.info("[{}] Consumer exiting main loop".format(self.trigger))
+            logging.info("[{}] Consumer exiting main "
```
bjustin-ibm: What is the reason for breaking this string?

mrutkows:
@bjustin These are all changes to adhere to PEP8 (<80 char lines and other things) that may appear "cosmetic" but to the Python community makes a significant different in terms of "correctness" to Python agreed-upon standards.

provider/app.py
mrutkows: Yes, it is more correct to test truth (boolean) in this manner in Python.

provider/app.py
mrutkows: @bjustin-ibm we are adding python PEP8 linting for all python files over time; Adhering to the <80 char line limit will be mandatory.

provider/consumer.py
mrutkows: @bjustin-ibm  Single quotes for symbolic strings, double quotes for (potentially) interpolated (or language translatable) strings.

bjustin-ibm:
> These are all changes to adhere to PEP8 (<80 char lines and other things) that may appear "cosmetic" but to the Python community makes a significant different in terms of "correctness" to Python agreed-upon standards.

I respectfully disagree about the line length. A quick survey of Stackoverflow shows that the 79 character limit imposed by PEP8 is extremely contentious and hotly debated. 

- https://stackoverflow.com/a/17319775
- https://stackoverflow.com/a/2793525

If you read PEP8 itself, they not only recognize this contention, but also explain that it is done only for readability on legacy displays which are only capable of displaying content 80 characters wide. It also allows for development teams to agree to enforce a longer line length: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

As a development team, I think we can agree that we are not developing this code on legacy displays that are incapable of displaying more than 80 characters and _also_ incapable of soft-wrapping long lines. I contend that the string breaking changes included in this PR actually reduce the readability of the code, rather than enhance it - e.g. [this proposed change](https://github.com/apache/incubator-openwhisk-package-kafka/pull/149/files/d91656bd7a3e5f68384489e998132efc99fe7966#diff-c7f0a1458defdca4247d971225ee27baR131).

I welcome a standard code formatting so long as it can be automated, and does not detract from the actual readability of the code by imposing an arcane limitation. 

I have some concrete suggestions for how to proceed:
- Please include a `pep8.rc` file that matches what scanCode expects. This will give us awareness of what needs to change without waiting for Travis, as well as the ability to configure our development tools to automatically fix the style issues.
- In that config file, let's set the line limit to either 120 or 160 characters. This will bring some amount of rigor and conformity to our formatting, without necessarily decreasing the readability.

rabbah:
+1 - we discussed this on the other repo and decided to not stick to the 80 chars per line. It really makes the code unreadable and as Justin pointed out these are plenty of reasons not to do this.

provider/app.py
rabbah: let's back off "mandatory" - we relaxed this 80 char limit in the main repo iirc.

mrutkows:
@bjustin-ibm will resubmit without any of the PEP8 style guide line length changes.

PR #151, opened 2017-03-15
Use CouchDB view instead of retrieving all docs

bjustin-ibm:
This strategy is more robust than fetching all docs and just assuming they represent triggers.

This change has a couple pieces:
- during DB migration, if needed, add a view that only contains valid trigger docs
- in Database.triggers() use the new view to retrieve just the triggers

This approach works in both cases where the DB does and does not already exist.


bjustin-ibm:
FYI @jeremiaswerner

csantanapr:
@jasonpet this solution is simple enough that logic should work for cloudant and alarm

PR #152, opened 2017-03-20
Verify migration

bjustin-ibm:
This PR has two parts:
- A manual script to verify the DB migration process. This script verifies that the service provider can create a DB from scratch with the required view for retrieving triggers, as well as add the view to an existing DB. It only needs to be run once (I've run it).
- Cloudant was being very, _very_ slow as I was developing this script, and I noticed that when this happens, the situation can arise where the database migration is kicked off before the DB has actually been created. For this I have moved Database initialization to inside the `main` function.

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -27,7 +27,7 @@ from service import Service
 app = Flask(__name__)
 app.debug = False
 
-database = Database()
+database = None
 consumers = ConsumerCollection()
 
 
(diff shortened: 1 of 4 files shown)


csantanapr:
@bjustin-ibm do you want to merge this PR or close it?

provider/app.py
```diff
@@ -231,6 +231,8 @@ def main():
     enable_generic_kafka = (generic_kafka == 'True')
     logging.info('enable_generic_kafka is {} {}'.format(enable_generic_kafka, type(enable_generic_kafka)))
 
+    global database
```
csantanapr: This changes still compatible with latest code in master?

bjustin-ibm: Yes I believe so.

csantanapr:
There is a conflict now

bjustin-ibm:
@csantanapr I have rebased this work and all tests still pass.

PR #154, opened 2017-03-28
Use Python 3 for Producer Feeds

dubeejw:
- Feeds actions fail now that the Python 3 is the default version for OpenWhisk actions

### DIFF
--- a/action/kafkaProduce.py
+++ b/action/kafkaProduce.py
@@ -1,4 +1,5 @@
 import ssl
+import base64
 from kafka import KafkaProducer
 from kafka.errors import NoBrokersAvailable
 
@@ -16,18 +17,18 @@ def main(params):
             api_version_auto_timeout_ms=15000,
             bootstrap_servers=brokers)
 
-        print "Created producer"
+        print("Created producer")
 
         # only use the key parameter if it is present
         if 'key' in validatedParams:
             messageKey = validatedParams['key']
-            producer.send(validatedParams['topic'], bytes(validatedParams['value']), key=bytes(messageKey))
+            producer.send(validatedParams['topic'], bytes(validatedParams['value'], 'utf-8'), key=bytes(messageKey, 'utf-8'))
         else:
-            producer.send(validatedParams['topic'], bytes(validatedParams['value']))
+            producer.send(validatedParams['topic'], bytes(validatedParams['value'], 'utf-8'))
 
         producer.flush()
 
-        print  "Sent message"
+        print("Sent message")
     except NoBrokersAvailable:
         # this exception's message is a little too generic
         return {'error': 'No brokers available. Check that your supplied brokers are correct and available.'}
@@ -39,8 +40,6 @@ def main(params):
 def validateParams(params):
     validatedParams = params.copy()
     requiredParams = ['brokers', 'topic', 'value']
-    actualParams = params.keys()
-
     missingParams = []
 
     for requiredParam in requiredParams:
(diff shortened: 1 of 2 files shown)


dubeejw:
@jasonpet, please review.

dubeejw:
@jasonpet, need to fix encoding still.

dubeejw:
@csantanapr, @jasonpet, tests are passing locally for me with this update, please review.

action/kafkaProduce.py
```diff
@@ -51,19 +50,21 @@ def validateParams(params):
...
 
     if 'base64DecodeValue' in params and params['base64DecodeValue'] == True:
-        decodedValue = params['value'].decode('base64').strip()
-        if len(decodedValue) == 0:
+        try:
+            validatedParams['value'] = base64.b64decode(params['value']).decode('utf-8')
+
+            if len(validatedParams['value']) == 0:
+                raise Exception
+        except:
```
jtyberg: It's typical in Python to be explicit in catching and throwing Exceptions.  Here it could be `binascii.Error`, `KeyError`, or decoding error.    Here you're also raising an Exception in the same block you're catching it.  I would split this block into try/catch for decoding, and separate the conditional check for the key being empty.  I think it would be more readable/clear that way.

action/kafkaProduce.py
```diff
@@ -1,4 +1,5 @@
 import ssl
+import base64
 from kafka import KafkaProducer
 from kafka.errors import NoBrokersAvailable
```
dubeejw: @jtyberg, unfortunately the updates are not compatible with Python 2 due to lines like this `bytes(validatedParams['value'], 'utf-8')`. The `bytes` method only takes one argument in Python 2.

jtyberg: @dubeejw We just went through this Python 2/3 bytes/text mess in actionProxy.  I think we can make this code work for both versions of Python.

In Python 2, `bytes` is a subset of `str`, and when you decode a `bytes/str` object, you get a unicode string. 

```
Python 2.7.13
>>> import base64
>>> winter64 = '4p2EIOKYgyDinYQgYnJyCg=='
>>> winter = base64.b64decode(winter64)
>>> winter
'\xe2\x9d\x84 \xe2\x98\x83 \xe2\x9d\x84 brr\n'
>>> type(winter)
<type 'str'>
>>> isinstance(winter, bytes)
True
>>> isinstance(winter, str)
True
>>> x = winter.decode('utf-8')
>>> type(x)
<type 'unicode'>
>>> isinstance(x, bytes)
False
>>> x
u'\u2744 \u2603 \u2744 brr\n'
>>> print(x)
❄ ☃ ❄ brr
```

In Python 3, `bytes` is binary data, `str` is a unicode string, and there is no `unicode` type.   So the types are a bit different, but the same code will give you the same output.

```
Python 3.5.4
>>> import base64
>>> winter64 = '4p2EIOKYgyDinYQgYnJyCg=='
>>> winter = base64.b64decode(winter64)
>>> winter
b'\xe2\x9d\x84 \xe2\x98\x83 \xe2\x9d\x84 brr\n'
>>> type(winter)
<class 'bytes'>
>>> isinstance(winter, bytes)
True
>>> isinstance(winter, str)
False
>>> x = winter.decode('utf-8')
>>> type(x)
<class 'str'>
>>> isinstance(x, bytes)
False
>>> x
'❄ ☃ ❄ brr\n'
>>> print(x)
❄ ☃ ❄ brr
```

It looks like the program is receiving base64 encoded strings, decoding to bytes, decoding the bytes to unicode strings, and finally, encoding the strings back to bytes before sending across the wire.

If the bytes to text conversion is needed, this works the same in both versions of Python:

```
>>> import base64
>>> winter64 = '4p2EIOKYgyDinYQgYnJyCg=='
>>> winter = base64.b64decode(winter64)
>>> x = winter.decode('utf-8')
>>> bytes(x.encode('utf-8'))
b'\xe2\x9d\x84 \xe2\x98\x83 \xe2\x9d\x84 brr\n'
```

If decoding bytes to text and back to bytes is not needed, then this works in both versions of Python:

```
>>> import base64
>>> winter64 = '4p2EIOKYgyDinYQgYnJyCg=='
>>> winter = base64.b64decode(winter64)
>>> bytes(winter)
b'\xe2\x9d\x84 \xe2\x98\x83 \xe2\x9d\x84 brr\n'
```

I know, it's clear as mud.  Basically, do you need to use the `decode` after `b64decode`?  If so, just invoke `encode` on the string object before sending to `bytes()`.

mrutkows:
Apart from your intent to preserve Python 2 and 3 with base64 encoding (decoding).  It would be good to update your files to be PEP8 compliant and Apache compliant as well.  

For the Apache (ASF) header, please see the attached file for format.

Your main issues for PEP8 compliance are using (exactly) 2 blank lines before any "def" (function declaration) including "main".   I will provide other comments in-line.

PR #160, opened 2017-04-13
HA - redundancy

bjustin-ibm:
This change introduces redundancy in the feed provider, allowing multiple instances to be deployed simultaneously such that if any one instance stops working, the remaining instances will automatically take over the work of the failed instance.

This is accomplished by changing the feed actions to write directly to the trigger DB, rather than talking directly to the feed provider through REST endpoints. In this way, provider instances become aware of new and deleted triggers by monitoring the DB changes feed. By using the same Kafka group id for every trigger, this ensures that only one provider instance will consume any produced message and therefore, only one will fire the trigger. Additionally, automatic failover and even a bit of load balancing among provider instances is automatically provided by Kafka rebalancing.

As a result, large amounts of code are no longer needed (REST endpoints, safety/security checks, validation, etc.) and have been removed, greatly simplifying the implementation.

### DIFF
--- a/action/kafkaFeed.js
+++ b/action/kafkaFeed.js
@@ -1,143 +1,30 @@
-var request = require('request');
+const common = require('./lib/common');
 
 /**
  *   Feed to listen to Kafka messages
  *  @param {string} brokers - array of Kafka brokers
  *  @param {string} topic - topic to subscribe to
  *  @param {bool}   isJSONData - attempt to parse messages as JSON
- *  @param {bool}   isBinaryKey - encode message as Base64
- *  @param {bool}   isBinaryValue - encode key as Base64
- *  @param {string} endpoint - address to OpenWhisk deployment
+ *  @param {bool}   isBinaryKey - encode key as Base64
+ *  @param {bool}   isBinaryValue - encode message as Base64
+ *  @param {string} endpoint - address to OpenWhisk deployment (expected to be bound at deployment)
  */
 function main(params) {
-    var promise = new Promise(function(resolve, reject) {
-        if(!params.package_endpoint) {
-            reject('Could not find the package_endpoint parameter.');
-            return;
-        }
+    const endpoint = params.endpoint;
+    const webActionName = 'kafkaFeedWeb';
 
-        var triggerComponents = params.triggerName.split("/");
-        var namespace = encodeURIComponent(process.env['__OW_NAMESPACE']);
-        var trigger = encodeURIComponent(triggerComponents[2]);
-        var feedServiceURL = 'http://' + params.package_endpoint + '/triggers/' + namespace + '/' + trigger;
+    var massagedParams = common.massageParamsForWeb(params);
+    massagedParams.triggerName = common.getTriggerFQN(params.triggerName);
 
-        if (params.lifecycleEvent === 'CREATE') {
-            // makes a PUT request to create the trigger in the feed service provider
-            var putTrigger = function(validatedParams) {
-                var body = validatedParams;
+    if (params.lifecycleEvent === 'CREATE') {
+        return common.createTrigger(endpoint, massagedParams, webActionName);
+    } else if (params.lifecycleEvent === 'DELETE') {
+        return common.deleteTrigger(endpoint, massagedParams, webActionName);
+    }
 
-                // params.endpoint may already include the protocol - if so,
-                // strip it out
-                var massagedAPIHost = params.endpoint.replace(/https?:\/\/(.*)/, "$1");
-                body.triggerURL = 'https://' + process.env['__OW_API_KEY'] + "@" + massagedAPIHost + '/api/v1/namespaces/' + namespace + '/triggers/' + trigger;
-
-                var options = {
-                    method: 'PUT',
-                    url: feedServiceURL,
-                    body: JSON.stringify(body),
-                    headers: {
-                        'Content-Type': 'application/json',
-                        'User-Agent': 'whisk'
-                    }
-                };
-
-                return doRequest(options);
-            };
-
-            validateParameters(params)
-                .then(putTrigger)
-                .then(resolve)
-                .catch(reject);
-        } else if (params.lifecycleEvent === 'DELETE') {
-            var authorizationHeader = 'Basic ' + new Buffer(process.env['__OW_API_KEY']).toString('base64');
-
-            var options = {
-                method: 'DELETE',
-                url: feedServiceURL,
-                headers: {
-                    'Content-Type': 'application/json',
-                    'Authorization': authorizationHeader,
-                    'User-Agent': 'whisk'
-                }
-            };
-
-            doRequest(options)
-                .then(resolve)
-                .catch(reject);
-        }
-    });
-
-    return promise;
+    return {
+        error: 'unsupported lifecycleEvent'
+    };
 }
 
-function doRequest(options) {
-    var requestPromise = new Promise(function (resolve, reject) {
-        request(options, function (error, response, body) {
-            if (error) {
-                reject({
-                    response: response,
-                    error: error,
-                    body: JSON.parse(body)
-                });
-            } else {
-                console.log("Status code: " + response.statusCode);
-
-                if (response.statusCode >= 400) {
-                    console.log("Response from Kafka feed service: " + body);
-                    reject({
-                        statusCode: response.statusCode,
-                        response: JSON.parse(body)
-                    });
-                } else {
-                    resolve({
-                        response: JSON.parse(body)
-                    });
-                }
-            }
-        });
-    });
-
-    return requestPromise;
-}
-
-function validateParameters(rawParams) {
-    var promise = new Promise(function(resolve, reject) {
-        var validatedParams = {};
-
-        validatedParams.isMessageHub = false;
-        validatedParams.isJSONData = getBooleanFromArgs(rawParams, 'isJSONData');
-        validatedParams.isBinaryValue = getBooleanFromArgs(rawParams, 'isBinaryValue');
-        validatedParams.isBinaryKey = getBooleanFromArgs(rawParams, 'isBinaryKey');
-
-        if (validatedParams.isJSONData && validatedParams.isBinaryValue) {
-            reject('isJSONData and isBinaryValue cannot both be enabled.');
-            return;
-        }
-
-        if (rawParams.topic && rawParams.topic.length > 0) {
-            validatedParams.topic = rawParams.topic;
-        } else {
-            reject('You must supply a "topic" parameter.');
-            return;
-        }
-
-        if (isNonEmptyArray(rawParams.brokers)) {
-            validatedParams.brokers = rawParams.brokers;
-        } else {
-            reject('You must supply a "brokers" parameter as an array of Kafka brokers.');
-            return;
-        }
-
-        resolve(validatedParams);
-    });
-
-    return promise;
-}
-
-function getBooleanFromArgs(args, key) {
-  return (typeof args[key] !== 'undefined' && args[key] && (args[key] === true || args[key].toString().trim().toLowerCase() === 'true'));
-}
-
-function isNonEmptyArray(obj) {
-    return obj && Array.isArray(obj) && obj.length !== 0;
-}
+exports.main = main;
--- a/action/kafkaFeedWeb_package.json
+++ b/action/kafkaFeedWeb_package.json
@@ -1,0 +1,5 @@
+{
+  "name": "kafkaFeedWeb",
+  "version": "1.0.0",
+  "main": "kafkaFeedWeb.js"
+}
--- a/action/kafkaFeed_package.json
+++ b/action/kafkaFeed_package.json
@@ -1,0 +1,5 @@
+{
+  "name": "kafkaFeed",
+  "version": "1.0.0",
+  "main": "kafkaFeed.js"
+}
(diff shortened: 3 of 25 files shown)


bjustin-ibm:
@jasonpet This is still WIP but currently quite functional. You may be interested in certain pieces that also apply to the work you are doing - especially `provider/Service.py` which handles the DB changes feed, and the `installCatalog.sh` script.

action/kafkaFeed.js
```diff
@@ -1,141 +1,208 @@
...
+    return url;
+}
+
+function ensureTriggerIsUnique(db, triggerFQN) {
+    return db.get(triggerFQN)
+        .then(result => {
+            return Promise.reject('Trigger already exists');
+        })
+        .catch(err => {
+            // turn that frown upside-down!
```
dubeejw: Why exactly?

installKafka.sh
```diff
@@ -34,16 +34,18 @@ fi
...
-KAFKA_PROVIDER_ENDPOINT=$KAFKA_TRIGGER_HOST':'$KAFKA_TRIGGER_PORT
-
 PACKAGE_HOME="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 export WSK_CONFIG_FILE= # override local property file to avoid namespace clashes
 
-echo Installing the kafka feed action.
+echo Installing the Kafka package and feed action.
 
 $WSK_CLI -i --apihost "$EDGEHOST" action update messaging/kafkaFeed "$PACKAGE_HOME/action/kafkaFeed.js" \
```
dubeejw: Make web action here.

installCatalog.sh
```diff
@@ -34,24 +34,24 @@ fi
...
     --auth "$AUTH" \
     --shared yes \
     -a parameters '[ {"name":"kafka_brokers_sasl", "required":true, "description": "Array of Message Hub brokers", "bindTime":true},{"name":"user", "required":true, "description": "Message Hub username", "bindTime":true},{"name":"password", "required":true, "description": "Message Hub password", "bindTime":true, "type":"password"},{"name":"topic", "required":true, "description": "Topic to subscribe to"},{"name":"isJSONData", "required":false, "description": "Attempt to parse message value as JSON"},{"name":"isBinaryKey", "required":false, "description": "Encode key as Base64"},{"name":"isBinaryValue", "required":false, "description": "Encode message value as Base64"},{"name":"endpoint", "required":true, "description": "Hostname and port of OpenWhisk deployment"},{"name":"kafka_admin_url", "required":true, "description": "Your Message Hub admin REST URL", "bindTime":true}]' \
     -p bluemixServiceName 'messagehub' \
-    -p endpoint "$APIHOST" \
-    -p package_endpoint $KAFKA_PROVIDER_ENDPOINT
+    -p endpoint "$APIHOST"
 
 $WSK_CLI -i --apihost "$EDGEHOST" action update messaging/messageHubFeed "$PACKAGE_HOME/action/messageHubFeed.js" \
```
dubeejw: This needs to be a web action per our discussion.

provider/app.py
```diff
@@ -14,133 +14,22 @@
...
-            response = jsonify({'error': 'not authorized'})
-            response.status_code = 401
-    else:
-        response = jsonify({'error': 'not found'})
-        response.status_code = 404
-
-    return response
-
-
 @app.route('/')
```
dubeejw: Do we really need this?

provider/app.py
```diff
@@ -150,72 +39,21 @@ def testRoute():
...
-        logging.info("Required fields are missing {}".format(missing))
-
-    return missing
-
-
 def main():
     logger = logging.getLogger()
     logger.setLevel(logging.INFO)
 
+    component = os.getenv('INSTANCE', 'messageHubTrigger-0')
```
dubeejw: Explain.

provider/consumer.py
```diff
@@ -207,7 +207,6 @@ def run(self):
 
         if self.desiredState() == Consumer.State.Dead:
             logging.info('[{}] Permanently killing consumer because desired state is Dead'.format(self.trigger))
-            self.database.deleteTrigger(self.trigger)
```
dubeejw: Assuming delete is done through an action now.

provider/app.py
dubeejw: Assuming this is for multiple instances of the provider.

provider/database.py
```diff
@@ -40,45 +40,8 @@ class Database:
         logging.warn('Database does not exist - creating it.')
         database = client.create_database(dbname)
 
-    def recordTrigger(self, triggerFQN, doc):
-        with self.lock:
```
dubeejw: Lock isn't being used anymore, so remove `lock = Lock()` line 29.

provider/service.py
```diff
@@ -0,0 +1,59 @@
...
+                        # is a valid trigger doc that has changed
+                        if 'triggerURL' in change['doc']:
+                            self.createAndRunConsumer(change['doc'])
+                        else:
+                            logging.debug('[changes] Found a change for a non-trigger document')
+                    else:
+                        # this isn't really supported just yet...
+                        logging.warn("[{}] Found a document update to an existing trigger... that's unexpected".format(change['id']))
+
+            logging.error("[changes] uh-oh! I made it out of the changes for loop!")
```
dubeejw: Uh oh, restart the container?

bjustin-ibm: Because the changes feed is "infinite" the `for` loop will keep going indefinitely even when there are no changes at the moment. Which is to say, line 52 is only ever executed when something unexpected happens. That's why it is logged as an error. The outer `while True` loop will automatically restart checking the changes feed.

provider/consumer.py
bjustin-ibm: Correct. Provider becomes aware of deletes through the changes feed.

provider/app.py
bjustin-ibm: Exactly. Done for logging so you can tell which instance generated the log message.

provider/app.py
bjustin-ibm: Sort of. It is a nice "ping" check to see if the provider is running at all.

installKafka.sh
bjustin-ibm: Yes, maybe. It is still being debated.

installCatalog.sh
bjustin-ibm: Probably 😃

installKafka.sh
```diff
@@ -34,16 +34,18 @@ fi
...
     --auth "$AUTH" \
+    -p CLOUDANT_USER "$DB_USER" \
+    -p CLOUDANT_PASS "$DB_PASS" \
+    -p CLOUDANT_DB "$DB_NAME" \
+    -a feed true \
     -a description 'Feed to listen to Kafka messages' \
     -a parameters '[ {"name":"brokers", "required":true, "description": "Array of Kafka brokers"}, {"name":"topic", "required":true, "description": "Topic to subscribe to"}, {"name":"isJSONData", "required":false, "description": "Attempt to parse message value as JSON"}, {"name":"isBinaryKey", "required":false, "description": "Encode key as Base64"}, {"name":"isBinaryValue", "required":false, "description": "Encode message value as Base64"}, {"name":"endpoint", "required":true, "description": "Hostname and port of OpenWhisk deployment"}]' \
```
bjustin-ibm: hmm... I'm a little surprised to see `endpoint` included. I would argue that parameters that are expected to be bound at deployment shouldn't be documented in the parameters annotation because it isn't really up to the user to supply them. Instead these should be documented in our build/deploy doc.

action/kafkaFeed.js
```diff
@@ -1,141 +1,211 @@
-var request = require('request');
+var request = require('request-promise');
+var Cloudant = require('cloudant');
```
markusthoemmes: Can we use `nano` over Cloudant here?

markusthoemmes:
From your description: Do I read it correctly, that each instance will setup a consumer for each trigger in the system, but only 1 will fire it due to Kafka's consumer group behavior? If one of the instances dies, Kafka rebalancing will do the job of reassigning "jobs" to the different instances?

action/kafkaFeed.js
bjustin-ibm: I believe we could. I don't think we are using any Cloudant-specific features.

bjustin-ibm:
> From your description: Do I read it correctly, that each instance will setup a consumer for each trigger in the system, but only 1 will fire it due to Kafka's consumer group behavior? If one of the instances dies, Kafka rebalancing will do the job of reassigning "jobs" to the different instances?

That is correct. On top of that, if the topic in question has multiple partitions, Kafka will automatically assign each instance a subset of partitions in the topic, allowing even for a bit of load balancing (as far as firing triggers) between the two instances.

markusthoemmes:
Sounds smart. What is the cost of creating a consumer though (even if no messages sent). Our scale will be capped by the number of consumers one instance can have open at any time, right?

bjustin-ibm:
Yes, that is also correct. This PR is merely Step 1, and your question goes straight to Step 3: Profit! Once we have this in place, the next step is to allow for multiple redundant pairs to exist which evenly share the load of all the triggers. At that point, we should be able to add new "worker" pairs as needed.

bjustin-ibm:
@mt now using `nano` instead of `cloudant` Confirm it works with CouchDB as well.
@dubeejw Now with zipped feed actions and code reuse.

action/lib/Database.js
```diff
@@ -0,0 +1,76 @@
...
+                    reject(err);
+                } else {
+                    resolve(result);
+                }
+            });
+        });
+    };
+};
+
+// these don't belong here and should be moved to a common place
```
bjustin-ibm: oh... right.

bjustin-ibm: oh... right.

bjustin-ibm:
Resolves #96

bjustin-ibm:
Rebased to include latest changes from master. Now supports dynamically disabling/activating triggers by listening to the DB changes feed.

action/messageHubFeedWeb.js
```diff
@@ -0,0 +1,192 @@
...
+        validatedParams.isBinaryKey = common.getBooleanFromArgs(rawParams, 'isBinaryKey');
+        validatedParams.endpoint = common.massageEndpointParam(rawParams.endpoint);
+        validatedParams.authKey = rawParams.authKey;
+
+        const uuid = require('uuid');
+        validatedParams.uuid = uuid.v4();
+
+        console.log(`I made it this far! ${JSON.stringify(validatedParams, null, 2)}`);
+
+        common.resolveTriggerDetails(validatedParams.authKey, validatedParams.endpoint, rawParams.triggerName)
```
bjustin-ibm: The whole point of this is to make sure we have the correct namespace. In this action, the namespace env var will be the action owner's (e.g. whisk.system), not the namespace for the trigger being created. However, @jasonpet points out that the correct namespace _is_ available in the non-web feed action, and could be simply passed into the web action as a parameter value. Brilliant!

action/lib/Database.js
```diff
@@ -0,0 +1,63 @@
+// constructor for DB object - a thin, promise-loving wrapper around nano
+module.exports = function(dbURL, dbName) {
+    console.log(`dbUrl: ${dbURL}`);
+    console.log(`dbName: ${dbName}`);
```
bjustin-ibm: Probably should get rid of these...

bjustin-ibm:
@dubeejw This is ready for full review.

action/kafkaFeedWeb_package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "kafkaFeedWeb",
+  "version": "1.0.0",
+  "main": "kafkaFeedWeb.js",
+  "dependencies" : {
```
bjustin-ibm: meh... probably not

action/kafkaFeed_package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "kafkaFeed",
+  "version": "1.0.0",
+  "main": "kafkaFeed.js",
+  "dependencies" : {
```
bjustin-ibm: See above reply.

action/lib/common.js
```diff
@@ -0,0 +1,140 @@
...
+function getBooleanFromArgs(args, key) {
+    return (typeof args[key] !== 'undefined' && args[key] && (args[key] === true || args[key].toString().trim().toLowerCase() === 'true'));
+}
+
+function isNonEmptyArray(obj) {
+    return obj && Array.isArray(obj) && obj.length !== 0;
+}
+
+// Return the trigger FQN with the resolved namespace
+// The is required to avoid naming conflicts when using the default namespace "_"
```
bjustin-ibm: I'm thinking `s/The/This/`

provider/service.py
```diff
@@ -0,0 +1,90 @@
...
+                        if existingConsumer.desiredState() == Consumer.State.Disabled and self.__isTriggerDocActive(document):
+                            # disabled trigger has become active
+                            logging.info('[{}] Existing disabled trigger should become active'.format(change["id"]))
+                            self.createAndRunConsumer(document)
+                        elif existingConsumer.desiredState() == Consumer.State.Running and not self.__isTriggerDocActive(document):
+                            # running trigger should become disabled
+                            logging.info('[{}] Existing running trigger should become disabled'.format(change["id"]))
+                            existingConsumer.disable()
+                        else:
+                            # TODO make this debug
```
bjustin-ibm: !!

bjustin-ibm:
@dubeejw I've addressed all your comments, and the tests still pass!

bjustin-ibm:
@csantanapr ready to merge!

PR #166, opened 2017-04-20
Reuse the basic health test trigger, if it already exists

### LINKED ISSUE #163: Only Create Health Test Trigger when it does not Exist
After initially creating the health test trigger, do not delete it. Instead reuse it for later health tests.

bjustin-ibm:
Check if the health test trigger already exists, if it does, reuse it. If not, create it and do not delete it at the end of the test.

Added some additional logic to prevent cross-pollination of triggers being fired by other tests simultaneously running and producing messages to the same topic. Previously, this was prevented by having a trigger name that contained a unique timestamp. Now we poll for up to 5 activations that have occurred since just before the message was produced. Existing code already handled multiple activations, and rooted through them all to find the one that contains the expected message.

Resolves #163

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -16,6 +16,9 @@
 
 package system.health
 
+import java.time.Clock
+import java.time.Instant
+
 import system.utils.KafkaUtils
 
 import scala.concurrent.duration.DurationInt
@@ -29,6 +32,9 @@ import org.scalatest.junit.JUnitRunner
 
 import common.JsHelpers
 import common.TestHelpers
+import common.TestUtils.DONTCARE_EXIT
+import common.TestUtils.NOT_FOUND
+import common.TestUtils.SUCCESS_EXIT
 import common.Wsk
 import common.WskActorSystem
 import common.WskProps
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -67,33 +71,38 @@ class BasicHealthTest
...
+
+                // It takes a moment for the consumer to fully initialize.
+                println("Giving the consumer a moment to get ready")
+                Thread.sleep(consumerInitTime)
+            } else {
+                println(s"Trigger already exists, reusing it: $triggerName")
+            }
 
             retry({
+                val start = Instant.now(Clock.systemUTC())
```
markusthoemmes: Note that `Instant` is in UTC by default.

bjustin-ibm: Good to know, but I suppose it doesn't hurt to be explicit.

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -67,33 +71,38 @@ class BasicHealthTest
...
-                        "topic" -> topic.toJson))
-            }
+            val result = wsk.trigger.get(triggerName, expectedExitCode = DONTCARE_EXIT)
 
-            withActivation(wsk.activation, feedCreationResult, initialWait = 5 seconds, totalWait = 60 seconds) {
-                activation =>
-                    // should be successful
-                    activation.response.success shouldBe true
-            }
+            if(result.exitCode != 0) {
```
dubeejw: Non-zero exit code doesn't guarantee the trigger doesn't exist. An exit code of 148 does.

bjustin-ibm: done

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -67,33 +73,38 @@ class BasicHealthTest
...
-                        "topic" -> topic.toJson))
-            }
+            val result = wsk.trigger.get(triggerName, expectedExitCode = DONTCARE_EXIT)
 
-            withActivation(wsk.activation, feedCreationResult, initialWait = 5 seconds, totalWait = 60 seconds) {
-                activation =>
-                    // should be successful
-                    activation.response.success shouldBe true
-            }
+            if(result.exitCode == resourceMissingExitCode) {
```
dubeejw: Use NOT_FOUND from TestUtils.java

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -67,33 +73,38 @@ class BasicHealthTest
...
+                withActivation(wsk.activation, feedCreationResult, initialWait = 5 seconds, totalWait = 60 seconds) {
+                    activation =>
+                        // should be successful
+                        activation.response.success shouldBe true
+                }
+
+                // It takes a moment for the consumer to fully initialize.
+                println("Giving the consumer a moment to get ready")
+                Thread.sleep(consumerInitTime)
+            } else {
```
dubeejw: Trigger may not exit here. Only if a 0 exit code is received.

bjustin-ibm: Good catch

bjustin-ibm: Fixed with an assertion that the exitCode should be 0

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -109,7 +120,7 @@ class BasicHealthTest
                     }
 
                 println("Polling for activations")
-                val activations = wsk.activation.pollFor(N = 1, Some(triggerName), retries = 30)
+                val activations = wsk.activation.pollFor(N = 5, Some(triggerName), since = Some(start), retries = 30)
```
dubeejw: Why N = 5?

bjustin-ibm: It is a completely arbitrary number > 1

I picked 5 because it seemed like the largest reasonable number of activations that could be caused by simultaneous tests posting a message to this same topic. Nevermind that such messages would likely get batched into one invoke... 5 seemed sufficiently large.

bjustin-ibm: It is important to note that we could make it a ridiculously large number and the test would not be affected. We are really only interested in 1 activation, and so this polling will almost always timeout after 30 seconds trying to obtain whatever number of activations we set here. The important thing is that the number is greater than 1 to handle what should be a very rare scenario.

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -67,33 +72,39 @@ class BasicHealthTest
...
+                    activation =>
+                        // should be successful
+                        activation.response.success shouldBe true
+                }
+
+                // It takes a moment for the consumer to fully initialize.
+                println("Giving the consumer a moment to get ready")
+                Thread.sleep(consumerInitTime)
+            } else {
+                result.exitCode shouldBe(0)
```
markusthoemmes: Should we make this `result.exitCode shouldBe SUCCESS_EXIT` (you're using `NOT_FOUND` above already)

bjustin-ibm: sigh... I just won't learn :wink:  Fixed.

rabbah: yes. yes you will. :D

PR #168, opened 2017-04-21
Disable triggers instead of deleting them

### LINKED ISSUE #167: Improve error handling on trigger fire, do not delete trigger mark it paused instead
Currently the trigger is being deleted from the DB on certain error scenarios

We need to improve how do we handle each use case, I think there are 3 use cases when firing

1. Trigger has being removed from OpenWhisk, Controller is reachable and returns Not Found (404)
2. Errors that are not recoverable (wrong auth 403, etc..), then delete trigger from db
3. Errors that with a retry can be succesfully fire on a retry, but never delete the trigger from db, and instead mark the trigger doc in db "paused: true"
4. others?

paused state for triggers is something that could be use when redudancy providers is implemented

bjustin-ibm:
Resolves #167

### DIFF
--- a/provider/app.py
+++ b/provider/app.py
@@ -129,7 +129,13 @@ def deleteTrigger(namespace, trigger):
     consumer = consumers.getConsumerForTrigger(triggerFQN)
     if consumer != None:
         if authorizedForTrigger(auth, consumer):
-            consumer.shutdown()
+            if consumer.desiredState() == Consumer.State.Disabled:
+                # it's already disabled, just delete it
+                database.deleteTrigger(triggerFQN)
+                consumers.removeConsumerForTrigger(triggerFQN)
+            else:
+                consumer.shutdown()
+
             response = jsonify({'success': True})
         else:
             response = jsonify({'error': 'not authorized'})
--- a/provider/app.py
+++ b/provider/app.py
@@ -158,19 +158,20 @@ def authorizedForTrigger(auth, consumer):
 
 
 def createAndRunConsumer(triggerFQN, params, record=True):
-    if app.config['TESTING'] == True:
-        logging.debug("Just testing")
-    else:
-        # generate a random uuid for new triggers
-        if not 'uuid' in params:
-            params['uuid'] = str(uuid.uuid4())
+    # generate a random uuid for new triggers
+    if not 'uuid' in params:
+        params['uuid'] = str(uuid.uuid4())
 
+    if 'status' not in params or params['status']['active'] == True:
+        logging.info('{} Trigger was determined to be active, starting...'.format(triggerFQN))
         consumer = Consumer(triggerFQN, params)
         consumer.start()
         consumers.addConsumerForTrigger(triggerFQN, consumer)
+    else:
+        logging.info('{} Trigger was determined to be disabled, ignoring...'.format(triggerFQN))
 
-        if record:
-            database.recordTrigger(triggerFQN, params)
+    if record:
+        database.recordTrigger(triggerFQN, params)
 
 
 def restoreTriggers():
--- a/provider/app.py
+++ b/provider/app.py
@@ -168,13 +168,17 @@ def createAndRunConsumer(triggerFQN, params, record=True):
     if not 'uuid' in params:
         params['uuid'] = str(uuid.uuid4())
 
+    # Create a representation for this trigger, even if it is disabled
+    # This allows it to appear in /health as well as allow it to be deleted
+    # Creating this object is lightweight and does not initialize any connections
+    consumer = Consumer(triggerFQN, params)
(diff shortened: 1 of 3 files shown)


bjustin-ibm:
FYI @csantanapr @jasonpet

provider/app.py
```diff
@@ -158,19 +164,20 @@ def authorizedForTrigger(auth, consumer):
...
+    if not 'uuid' in params:
+        params['uuid'] = str(uuid.uuid4())
 
+    if 'status' not in params or params['status']['active'] == True:
+        logging.info('{} Trigger was determined to be active, starting...'.format(triggerFQN))
         consumer = Consumer(triggerFQN, params)
         consumer.start()
         consumers.addConsumerForTrigger(triggerFQN, consumer)
+    else:
+        logging.info('{} Trigger was determined to be disabled, ignoring...'.format(triggerFQN))
```
dubeejw: From here, recordTrigger() is called, which sets the disabled trigger's status to active.

bjustin-ibm: During provider boot up, this function is called for all existing triggers in the DB with the `record` flag False. That is, if the trigger already exists in the DB (currently all disabled triggers are already in the DB as this function is never called to create a new disabled trigger), then this code will not write that document again.

Conversely, you _could_ call this function as-is to create a new disabled trigger, in which case you would want to record it to the DB. Although this will never happen, the current code would correctly handle it.

bjustin-ibm: Regardless, perhaps I should add a check in `recordTrigger()` to set the `status` to active only if the `status` key does not already exist in the dictionary.

bjustin-ibm: ^^ done.

provider/database.py
```diff
@@ -46,6 +47,15 @@ def recordTrigger(self, triggerFQN, doc):
                 document = dict(doc)
                 document['_id'] = triggerFQN
 
+                if 'status' not in document:
+                    # set the status as active
+                    status = {
+                        'active': True,
+                        'dateChanged': time.time()
```
csantanapr: like the name `dateChanged`

provider/consumer.py
```diff
@@ -291,6 +294,12 @@ def __pollForMessages(self):
         self.updateLastPoll()
         return messages
 
+    # decide whether or not to disable a trigger based on the status code returned
+    # from firing the trigger. Specifically, disable on all 4xx status codes
+    # except 408 (gateway timeout) and 429 (throttle)
+    def __shouldDisable(self, status_code):
+        return status_code in range(400, 500) and status_code not in [408, 429]
```
csantanapr: This means, we disable for Not Found (404), Unauthorized (401) or Forbidden (403) which is think is correct, this can be clean at a later time by another external process.

bjustin-ibm:
@dubeejw I have one more potential change in progress for this. Currently, triggers that are already disabled at boot cannot be properly deleted (because they are not represented in the ConsumerCollection).

bjustin-ibm:
@dubeejw Done. Please see 8fbc26006109628c753ddfa7a6a2ac38bd96b0fa

provider/app.py
```diff
@@ -158,19 +164,24 @@ def authorizedForTrigger(auth, consumer):
...
+    if not 'uuid' in params:
+        params['uuid'] = str(uuid.uuid4())
+
+    # Create a representation for this trigger, even if it is disabled
+    # This allows it to appear in /health as well as allow it to be deleted
+    # Creating this object is lightweight and does not initialize any connections
+    consumer = Consumer(triggerFQN, params)
+    consumers.addConsumerForTrigger(triggerFQN, consumer)
+
+    if 'status' not in params or params['status']['active'] == True:
```
dubeejw: I don't think the `status` field will already exist when the container with this new code update is started for the first time. Will that result in all the existing triggers being disabled?

provider/app.py
```diff
@@ -158,19 +164,24 @@ def authorizedForTrigger(auth, consumer):
...
+        logging.info('{} Trigger was determined to be active, starting...'.format(triggerFQN))
         consumer.start()
-        consumers.addConsumerForTrigger(triggerFQN, consumer)
+    else:
+        logging.info('{} Trigger was determined to be disabled, not starting...'.format(triggerFQN))
```
dubeejw: If a consumer is disabled by accident, does the entire container have to be restarted in order to rename that consumer?

bjustin-ibm: Yes. This is temporary until #160 is finished. At that point, the provider will listen to DB changes directly and will be able to react to changes in status.

dubeejw: If the above is true, perhaps The Doctor could poll for consumers that should be reactivated.

bjustin-ibm: @dubeejw FYI: 93000edde4a7c05ae45956fa4997f16c2e838d6c

PR #175, opened 2017-05-11
fix logging timestamp format

bjustin-ibm:
Use a more standard format for the logging timestamp.


bjustin-ibm:
@dubeejw please review

PR #178, opened 2017-06-06
convert from threading to multiprocessing

bjustin-ibm:
Despite having a threading API, the Python GIL ensures that no Python code in the same process is allowed to execute in parallel. Threading, it seems, is a tool to work with concurrency without real parallel execution :( However, the `multiprocessing` module allows for true parallel execution by spawning separate Python processes. 

In order to accommodate more and more consumers, I have ported this over to use the `multiprocessing` module which, thankfully, has pretty much the same API as `threading`. Interprocess communication is accomplished by using shared dictionaries that can be read/written from the main process and the child consumer processes. Each shared dictionary conveys the current status of a single consumer, as well as serves as a vehicle for requesting state changes to the child process by simply setting the `desiredState` key in the shared dictionary.

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -213,6 +213,8 @@ class ConsumerThread (Thread):
                 if len(messages) > 0:
                     self.__fireTrigger(messages)
 
+                time.sleep(0.1)
+
             logging.info("[{}] Consumer exiting main loop".format(self.trigger))
         except Exception as e:
             logging.error('[{}] Uncaught exception: {}'.format(self.trigger, e))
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw please review

provider/consumer.py
```diff
@@ -213,6 +205,8 @@ def run(self):
                 if len(messages) > 0:
                     self.__fireTrigger(messages)
 
+                time.sleep(0.1)
```
dubeejw: Why pause here?

bjustin-ibm: I believe this may significantly reduce the CPU usage of an idle Consumer. Without a nice little pause, this loop is always running the CPU as hard as possible.

bjustin-ibm:
@csantanapr Please review/merge. @dubeejw has already given the 👍

provider/consumer.py
csantanapr: this smells

PR #180, opened 2017-06-13
Test fixes

bjustin-ibm:
Changes are to better avoid cross-contamination for trigger activations during automated tests in the scenario where the kafka instance being used by the tests is shared.

When the kafka instance under test is shared, it could be that messages are being produced to it from other sources. The automated tests should be aware of that and handle, as well as can be expected, to sift through potentially unexpected trigger activations that may result from other uses of the kafka instance.


bjustin-ibm:
@dubeejw please review

PR #186, opened 2017-06-23
automatically monitor and restore the health of the DB changes feed

### LINKED ISSUE #185: DB changes feed can become unrespsonsive - does not automatically recover
I observed a situation where the changes feed was no longer detecting changes for an extended period of time. I would have expected the DB client to automatically recover the feed, but it did not.

Solution could be to monitor the status of the changes feed and then automatically reinitialize it if it is found to be unresponsive.

bjustin-ibm:
Resolves #185 

The idea here is to periodically create non-trigger "canary" documents into the DB. The changes feed Service will monitor the time that the last canary was detected by the changes feed, and if the time exceeds a specified limit, it will automatically re-initialize the changes feed.

Additionally, I added a health test to ensure that new triggers could be successfully created. This required that the feed action return the UUID of the created trigger, so that it could be verified by the test through the `/health` endpoint.

### DIFF
--- a/action/lib/common.js
+++ b/action/lib/common.js
@@ -95,8 +95,10 @@ function createTrigger(endpoint, params, actionName) {
 
     return request(options)
         .then(response => {
-            // do not pass the response back to the caller, its contents are secret
-            return;
+            console.log(`response ${JSON.stringify(response, null, 2)}`)
+            return {
+                uuid: response
+            };
         })
         .catch(error => {
             console.log(`Error creating trigger: ${JSON.stringify(error, null, 2)}`);
--- a/provider/database.py
+++ b/provider/database.py
@@ -19,6 +19,7 @@ import uuid
 
 from cloudant.client import CouchDB
 from cloudant.result import Result
+from datetime import datetime
 
 class Database:
     db_prefix = os.getenv('DB_PREFIX', '')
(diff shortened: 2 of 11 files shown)


bjustin-ibm:
@jasonpet @dubeejw please review.  

For testing, I took a semi-manual approach. Locally, I set the `canaryTimeout` value well below the `canaryInterval` value, causing the changes feed to get reset automatically about every 30 seconds. I then ran the full automated test suite several times while the provider service was running this way.

jasonpet:
The node.js updates and the tests look good to me.  I will let @dubeejw handle the reviews for the python code.

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -69,8 +71,43 @@ class BasicHealthTest
...
     val kafkaUtils = new KafkaUtils
 
+    val getMessagingAddress =
+      if (System.getProperty("host") != "" && System.getProperty("port") != "") {
+        "http://" + System.getProperty("host") + ":" + System.getProperty("port")
+      }
+
     behavior of "Message Hub feed"
 
+    it should "create a new trigger" in withAssetCleaner(wskprops) {
```
dubeejw: Nice!

action/lib/common.js
```diff
@@ -95,8 +95,10 @@ function createTrigger(endpoint, params, actionName) {
 
     return request(options)
         .then(response => {
-            // do not pass the response back to the caller, its contents are secret
-            return;
+            console.log(`response ${JSON.stringify(response, null, 2)}`)
+            return {
+                uuid: response
```
dubeejw: 👍

provider/health.py
```diff
@@ -132,8 +133,9 @@ def getConsumers(consumers):
 
     return consumerReports
 
-def generateHealthReport(consumers):
+def generateHealthReport(consumers, lastCanaryTime):
     healthReport = {}
+    healthReport['last_db_canary'] = secondsSince(lastCanaryTime)
```
dubeejw: 👍

provider/service.py
```diff
@@ -13,62 +13,115 @@
...
+
+                        # delete the canary document
+                        self.database.deleteDoc(change['id'])
+                    else:
+                        logging.debug('[changes] Found a change for a non-trigger document')
 
+                if secondsSince(self.lastCanaryTime) > canaryTimeout:
+                    logging.warn('[canary] It has been more than {} seconds since the last canary - restarting the DB changes feed'.format(canaryTimeout))
+                    self.restartChangesFeed()
+                    break
```
dubeejw: This break is slightly confusing since the code block is so big. At first I thought you were jumping out of the main loop. Had to scroll up to the top of the function to see what was going on. Looks like you're just skipping the log message below.

dubeejw: How do we detect if an infinite loop is occurring here? Maybe add a counter for consecutive changes feed restarts.

provider/database.py
```diff
@@ -67,9 +68,42 @@ def disableTrigger(self, triggerFQN, status_code):
...
+            return self.database.infinite_changes(include_docs=True, heartbeat=(timeout*1000))
+        else:
+            return self.database.infinite_changes(include_docs=True, heartbeat=(timeout*1000), since=since)
+
+    def createCanary(self):
+        retryCount = 3
+
+        while retryCount >= 0:
+            if retryCount == 0:
+                logging.error('[canary] Retried and failed {} times to create a canary'.format(retryCount))
```
dubeejw: If canary document is not created by this function will the attempt to create it again occur when the canary is updated?

provider/service.py
bjustin-ibm: Let's say we do that... then what do we do with that information?

bjustin-ibm: I agree the break is confusing. I added a comment explaining what it does and why it is needed.

PR #191, opened 2017-06-30
Add ASF license to python files; no line limit changes.

mrutkows:
This PR seeks to replace PR https://github.com/apache/incubator-openwhisk-package-kafka/pull/149 with only addition of ASF headers and a small number of spacing issues.  No 79 char line limits changes included.

Any previous changes to account for proper use of "is, is not" for use with built-in constants True, False and None are removed.  However, we should look to fix these where possible as these may be more performant and perform the correct boolean comparison that was intended.

In addition, I did not attempt to adjust string literals to have single quotes and have translatable strings have double quotes.

### DIFF
--- a/action/kafkaProduce.py
+++ b/action/kafkaProduce.py
@@ -1,7 +1,28 @@
+"""Kafka message producer.
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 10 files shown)


bjustin-ibm:
Thanks @mrutkows!

PR #194, opened 2017-07-04
Fix infinite loop in error handling

bjustin-ibm:
Due to a logic error, the retry loop for creating a canary would loop infinitely after `retryCount` got to 0. Rework the broken logic to ensure the retry loop exits as expected.


bjustin-ibm:
@dubeejw please review.

PR #197, opened 2017-07-10
Use one canary document per provider instance

### LINKED ISSUE #189: harmless warning logs when deleting a non-existent canary
If you have multiple instances of the service running simultaneously, each becomes aware of all the canaries created by all the instances. As such, there are a lot of (actually harmless) warning messages about not being able to delete a canary, because some other instance already deleted it.

Either:
- stop logging this, or turn it into a "debug" loglevel
- ensure that each instance only deletes the canaries it creates. This could be done by adding a unique ID to each canary document so that the instance that produced it can identify it.

The second option seems more correct to me.

bjustin-ibm:
This eliminates canary document deletes, which is good news for the changes feed as it won't have to process all the deletes when the container starts. This also avoids the problem of getting a 404 while trying to delete a canary that was already deleted by another instance.

Resolves #189

### DIFF
--- a/provider/database.py
+++ b/provider/database.py
@@ -41,6 +41,9 @@ class Database:
     filters_design_doc_id = '_design/filters'
     only_triggers_view_id = 'only-triggers'
 
+    instance = os.getenv('INSTANCE', 'messageHubTrigger-0')
+    canaryId = "canary-{}".format(instance)
+
     if dbname in client.all_dbs():
         logging.info('Database exists - connecting to it.')
         database = client[dbname]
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw please review.

provider/database.py
```diff
@@ -83,33 +83,37 @@ def createCanary(self):
...
+        canaryId = "canary-{}".format(instance)
+
         while retryCount < maxRetries:
             try:
-                document = dict()
-                document['canary'] = datetime.now().isoformat()
+                if canaryId in self.database.keys(remote=True):
+                    # update the timestamp to cause a document change
+                    logging.debug("[database] Canary doc exists, updating it.")
+
```
dubeejw: Why a new line here, but not in the else block under the debug statement?

provider/database.py
```diff
@@ -83,33 +83,37 @@ def createCanary(self):
         maxRetries = 3
         retryCount = 0
 
+        instance = os.getenv('INSTANCE', 'messageHubTrigger-0')
```
dubeejw: Feels like the assignments to `instance` and `canaryId` should occur outside of this function. Otherwise, you are unnecessarily redefining them every time `createCanary()` is invoked.

bjustin-ibm: sounds reasonable

provider/database.py
bjustin-ibm: I'll put one after L101

bjustin-ibm:
@dubeejw I made updates based on your feedback. Please re-review.

bjustin-ibm:
@mrutkows James has reviewed this code and it is ready for merge.

mrutkows:
@houshengbo thanks Vincent

Issue #195, opened 2017-07-06
Do not re-use CouchDB client between processes


bjustin-ibm:
In a multiprocessing environment, SSL connections cannot readily be shared between processes. As the CouchDB client may use an SSL connection itself, it should not be shared between processes.

A fix would be to give Database.py a proper `__init__` function and require individual processes to instantiate their own DB client.  Additionally, the client should be instantiated as a instance field for all Process subclasses that require it.

PR #198, opened 2017-07-11
Do not attempt to share DB connections across multiple processes

bjustin-ibm:
The problem is that keeping the DB connection as a class property causes it to be re-used, potentially between multiple processes. For SSL connections in Python, this is not supported.

The fix is to ensure that the DB connection is made an instance field of the Database class, and also to ensure that all Database instances are themselves instance fields of their respective classes. This avoids sharing trying to share the same connection between multiple Python processes.

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -118,8 +118,6 @@ class ConsumerProcess (Process):
     retry_timeout = 1   # Timeout in seconds
     max_retries = 10    # Maximum number of times to retry firing trigger
 
-    database = Database()
-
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw please review fc47ac116c131c5a134d48a9a8983b4ced315b26 only.  I'll rebase once #197 is merged.

dubeejw:
@houshengbo, please merge.

bjustin-ibm:
Resolves #195

dubeejw:
@houshengbo, please merge.

PR #200, opened 2017-07-18
Add a 1 second delay between test retries

bjustin-ibm:
After creating a trigger, there may be some time before the service is aware of the new trigger. Add a 1 second delay to the retry attempts at verifying the trigger was created.


bjustin-ibm:
@dubeejw please review

dubeejw:
@csantanapr, please merge.

PR #205, opened 2017-07-25
Better handle when the connection to the DB drops

bjustin-ibm:
Make use of a timeout value when constructing the changes feed. This ensures that if the connection to the DB is lost, an exception is thrown in a timely manner so that the changes feed can be restarted. Without this timeout set on the database client, the changes feed will simply hang when the connection is lost, and may not recover.

Also, might as well update the DB client while we're at it...

### DIFF
--- a/provider/service.py
+++ b/provider/service.py
@@ -25,19 +25,14 @@ from consumer import Consumer
 from database import Database
 from datetime import datetime
 from datetimeutils import secondsSince
+from requests.exceptions import ConnectionError, ReadTimeout
 from threading import Thread
 
 # How often to produce canary documents
 canaryInterval = 60  # seconds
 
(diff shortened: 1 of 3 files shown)


bjustin-ibm:
@dubeejw please review

provider/service.py
```diff
@@ -58,76 +49,76 @@ def run(self):
...
+                self.stopChangesFeed()
 
             logging.debug("[changes] I made it out of the changes loop!")
 
-    def restartChangesFeed(self):
+    def stopChangesFeed(self):
         if self.changes != None:
             self.changes.stop()
 
         self.changes = None
```
dubeejw: This can be in the if block above.

dubeejw:
@csantanapr, merge.

PR #209, opened 2017-08-03
exponential backoff on trigger fire retry

### LINKED ISSUE #207: look into making backoff exponential vs. linear
today default backoff is 10 tries with 1 second delay.
We had feedback asking why not make this exponential.

bjustin-ibm:
Resolves #207 

With this algorithm, there are 6 retry attempts with a delay of `2^x` seconds between attempts. This has the following effects:
- Retries initially happen quickly
- Total possible retry length (from first failure to last attempt) is extended to 126 seconds
- The last retry delay is 64 seconds which will ensure that the 60 second rolling throttle window has passed between the last two attempts

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -114,9 +114,7 @@ class Consumer:
 
 
 class ConsumerProcess (Process):
-
-    retry_timeout = 1   # Timeout in seconds
-    max_retries = 10    # Maximum number of times to retry firing trigger
+    max_retries = 6    # Maximum number of times to retry firing trigger
 
     def __init__(self, trigger, params, sharedDictionary):
         Process.__init__(self)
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@dubeejw please review

provider/consumer.py
```diff
@@ -371,10 +369,11 @@ def __fireTrigger(self, messages):
                 if retry:
                     retry_count += 1
 
-                    if retry_count < self.max_retries:
+                    if retry_count <= self.max_retries:
+                        sleepyTime = pow(2,retry_count)
                         logging.info("[{}] Retrying in {} second(s)".format(
-                            self.trigger, self.retry_timeout))
-                        time.sleep(self.retry_timeout)
+                            self.trigger, sleepyTime))
```
dubeejw: Why a new line for this?

provider/thedoctor.py
```diff
@@ -27,7 +27,8 @@
 
 class TheDoctor (Thread):
     # maximum time to allow a consumer to not successfully poll() before restarting
-    poll_timeout_seconds = 20
+    # this value must be greater than the total amount of time a consumer might retry firing a trigger
+    poll_timeout_seconds = 200
```
dubeejw: Why 200 here? `2^1 + 2^2 + 2^3 + 2^4 + 2^5 + 2^6 = 126`

bjustin-ibm: This is the maximum amount of time between successful polls before the doctor will forcibly restart the consumer. It is a drastic measure that should not happen unless absolutely necessary. Keep in mind that previously, the total retry time was only 10 seconds and the `poll_timeout_seconds` was double that.

provider/consumer.py
bjustin-ibm: No idea...

bjustin-ibm:
@csantanapr please review/merge.

PR #214, opened 2017-08-30
Support multiple workers

abaruni:


abaruni:
@dubeejw would you mind giving a review of this PR?

dubeejw:
@abaruni, I'll take a look.

abaruni:
thanks

provider/database.py
```diff
@@ -37,9 +37,11 @@ class Database:
 
     filters_design_doc_id = '_design/filters'
     only_triggers_view_id = 'only-triggers'
+    by_worker_view_id = 'by-worker'
 
     instance = os.getenv('INSTANCE', 'messageHubTrigger-0')
-    canaryId = "canary-{}".format(instance)
+    workerId = os.getenv('WORKER', 'worker0')
```
dubeejw: Looks like the call to determine the worker occurs in service.py. Instead value should be determine once and shared between the two classes.

provider/service.py
```diff
@@ -120,6 +131,14 @@ def run(self):
 
             logging.debug("[changes] I made it out of the changes loop!")
 
+    def __isTriggerDocAssignedToMe(self, doc):
+        if "worker" in doc:
+            return doc["worker"] == self.workerId
+        else:
+            # TODO I'm uncomfortable with the assumption that "worker0" is the default worker
```
dubeejw: Address TODO.

dubeejw:
Need to look through this more. Are you using two views? @jasonpet, mentioned that it is possible to only use one view.

provider/database.py
```diff
@@ -37,6 +37,7 @@ class Database:
 
     filters_design_doc_id = '_design/filters'
     only_triggers_view_id = 'only-triggers'
+    by_worker_view_id = 'by-worker'
 
     instance = os.getenv('INSTANCE', 'messageHubTrigger-0')
     canaryId = "canary-{}".format(instance)
```
abaruni: I removed the worker id from the canary doc id. i originally added it to avoid conflict between multiple workers, but it is not needed as updating 'instance' for each additional worker would suffice

provider/service.py
```diff
@@ -120,6 +131,12 @@ def run(self):
 
             logging.debug("[changes] I made it out of the changes loop!")
 
+    def __isTriggerDocAssignedToMe(self, doc):
+        if "worker" in doc:
+            return doc["worker"] == self.workerId
+        else:
+            return self.workerId == "worker0"
```
abaruni: after speaking with @jasonpet this is the appropriate behavior. removed TODO comment

installKafka.sh
```diff
@@ -23,6 +23,7 @@ EDGEHOST="$2"
 DB_URL="$3"
 DB_NAME="${4}ow_kafka_triggers"
 APIHOST="$5"
+WORKERS="$6"
```
dubeejw: Need to update doc for this new parameter. See https://github.com/apache/incubator-openwhisk-package-kafka/blob/master/devGuide.md#install-actions.

action/lib/Database.js
```diff
@@ -60,4 +63,39 @@ module.exports = function(dbURL, dbName) {
                 });
             })
     };
+
+    this.getTriggerAssignment = function(workers) {
+        // a map between available workers and their number of assigned triggers
+        // values will be populated with the results of the assignment view
+        console.log(`Available workers: ${JSON.stringify(workers, null, 2)}`);
```
dubeejw: This log statement seems misleading since `worker0` is used if the workers parameter has an empty array as an argument.

dubeejw: Referring to this line below: `var assignment = workers[0] || 'worker0';`.

devGuide.md
```diff
@@ -45,6 +46,7 @@ Each script requires a number of arguments which are outlined below:
 |dburl|The full URL (including credentials) of the CouchDB or Cloudant account used by the feed service.|
 |dbprefix|A prefix to be prepended to the default DB name (ow_kafka_triggers) that will be created by the provider service.|
 |apihost|The hostname or IP address of the core OpenWhisk system that will be used as the hostname for all trigger URLs. In most cases, this will be the same as `edgehost`.|
+|workers|A comma separated list of the IDs of the running instances with each ID of the form `workerX`. e.g. `worker0,worker1`|
```
dubeejw: There is an example install script with arguments below. Needs to be updated with workers.

dubeejw:
Looks like Travis failed: ```
  [/home/travis/build/apache/incubator-openwhisk-package-kafka/tools/travis/../../tests/src/test/scala/system/packages/MessageHubMultiWorkersTest.scala]:
     193: file does not end with EOL.
Scan detected 1 error(s) in 1 file(s):
```

File: action/lib/Database.js [hunk 8]
```diff
@@ -60,4 +63,40 @@ module.exports = function(dbURL, dbName) {
                 });
             })
     };
+
+    this.getTriggerAssignment = function(workers) {
+        // a map between available workers and their number of assigned triggers
```
jasonpet: no need to do fetch the view and do all this work if available worker list is not > 1

bjustin-ibm:
🎉 Nice work @abaruni! 🎉

PR #216, opened 2017-10-02
update test to delete artifacts. reformat with scalariform

abaruni:
the second test in BasicHealthTest.scala was not deleting the test trigger after it completed. this PR is to ensure that test cleans up test artifacts after completing. i've also reformatted the class to use the recommended white spacing

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -47,141 +47,146 @@ import com.jayway.restassured.RestAssured
 
 import whisk.utils.retry;
 
-
 @RunWith(classOf[JUnitRunner])
 class BasicHealthTest
-    extends FlatSpec
-    with Matchers
-    with WskActorSystem
-    with BeforeAndAfterAll
-    with TestHelpers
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -47,141 +47,146 @@ import com.jayway.restassured.RestAssured
...
 
-    val messagingPackage = "/whisk.system/messaging"
-    val messageHubFeed = "messageHubFeed"
-    val messageHubProduce = "messageHubProduce"
+      if (result.exitCode == NOT_FOUND) {
+        // trigger does not yet exist, create it
+        println(s"Creating trigger ${triggerName}")
 
-    val consumerInitTime = 10000 // ms
+        val feedCreationResult = assetHelper.withCleaner(wsk.trigger, triggerName) {
```
abaruni: this is the actual code change to ensure this test deletes the artifacts after it completes. everything else is just reformatting from scalariform

abaruni:
@csantanapr would you mind taking a look at this?

PR #217, opened 2017-10-11
Add ability to get trigger configuration and status

abaruni:


abaruni:
@jasonpet @csantanapr mind having a look at what's here?

csantanapr:
@jasonpet can you review this?

action/kafkaFeedWeb.js
```diff
@@ -61,6 +61,35 @@ function main(params) {
 
                     resolve(common.webResponse(statusCode, body));
                 });
+        } else if (params.__ow_method === "get") {
```
dubeejw: Code is the same as in messageHubFeedWeb.js. Should make it a shared method between the two.

action/kafkaFeedWeb.js
```diff
@@ -61,6 +61,35 @@ function main(params) {
...
+                            triggerName: triggerDoc.triggerName,
+                            topic: triggerDoc.topic,
+                            isJSONData: triggerDoc.isJSONData,
+                            isBinaryValue: triggerDoc.isBinaryValue,
+                            isBinaryKey: triggerDoc.isBinaryKey,
+                            isMessageHub: triggerDoc.isMessageHub,
+                            brokers: triggerDoc.brokers,
+                            kafka_admin_url: triggerDoc.kafka_admin_url,
+                            username: triggerDoc.username,
+                            password: triggerDoc.password
```
dubeejw: Do we really want to return the password?

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
```diff
@@ -274,6 +274,67 @@ class MessageHubFeedTests
...
+        "password" -> password,
+        "api_key" -> kafkaUtils.getAsJson("api_key"),
+        "kafka_admin_url" -> admin_url,
+        "kafka_brokers_sasl" -> brokers,
+        "topic" -> topic.toJson,
+        "isBinaryKey" -> false.toJson,
+        "isBinaryValue" -> false.toJson
+      ))
+
+      val run = wsk.action.invoke(actionName, parameters = Map(
```
dubeejw: Should have a test to ensure this fails with an bad auth key.

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
```diff
@@ -274,6 +274,67 @@ class MessageHubFeedTests
...
+      withActivation(wsk.activation, run) {
+        activation =>
+          activation.response.success shouldBe true
+
+          inside (activation.response.result) {
+            case Some(result) =>
+              val config = result.getFields("config").head.asInstanceOf[JsObject].fields
+              val status = result.getFields("status").head.asInstanceOf[JsObject].fields
+
+              config should contain("brokers" -> brokers)
```
dubeejw: You should be able build a JsObject that has all these fields then do one or two comparisons instead of several.

dubeejw: Ex:
```
val expectedConfig = JsObject("isBinaryKey" -> false.toJson, "isBinaryValue" -> false.toJson, etc...)
config shouldBe(expectedConfig)
```

action/lib/common.js
```diff
@@ -130,6 +130,30 @@ function deleteTrigger(endpoint, params, actionName) {
         });
 }
 
+function getTrigger(endpoint, params, actionName) {
+    var options = {
+        method: 'GET',
+        url: getWebActionURL(endpoint, actionName),
+        rejectUnauthorized: false,
```
dubeejw: What's `rejectUnauthorized` do?

action/kafkaFeedWeb.js
abaruni: you'll notice the code between messageHubFeedWeb.js and kafkaFeedWeb.js is entirely duplicated. i think it's definitely fair to get rid of the duplication. but i don't this this is the the place to do it. it should be done under a different issue/epic

action/kafkaFeedWeb.js
abaruni: at this point in time yes. the user needs to be authenticated. this is no different than fetching the service credentials for your instance of message hub from bluemix

abaruni: we will likely add flags in the future to adjust what is returned. @csantanapr wants to get this initial work out before focusing on that as this is highly desirable

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: that would work except for the fact the trigger name as returned is not equal to the trigger name passed in

abaruni: and i would still have to test fields individually for status as i cannot know the value of the dateChanged field

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: shouldn't that be a test for testing the action itself? i.e. failing with a bad auth key is functionality that is inherent to invoking the action itself, not invoking the action with the parameter `lifecycleEvent` set to `READ`. i'm all for having this test if it doesn't exist for the provider already. but a separate issue should be opened as it doesn't specifically relate to this functionality. i.e. an action should never be able to be invoked without proper authKey

action/lib/common.js
abaruni: not entirely sure. i think it might have something to do with accepting self-signed certs https://github.com/request/request-promise/issues/225

abaruni: not entirely sure though

action/kafkaFeedWeb.js
dubeejw: The code is not entirely duplicated. messageHubFeedWeb.js has a different method to validate parameters. Most of the shared code there is separated out into functions that is shared amongst the new files.

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
dubeejw: I mean the authKey you are passing as a parameter to the action.

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
dubeejw: What does the triggerName look like when it is returned here?

action/kafkaFeedWeb.js
dubeejw: @csantanapr, thoughts on exposing the password here?

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: passed in as `/_/dummyTrigger-...` returned as `/guest/dummyTrigger-...`

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: right. that is a required parameter for the action itself regardless of the value of `lifecycleEvent`

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
dubeejw: So we know what the returned trigger name will be. That means the config property can be compared to a JsObject.

What does dateChanged look like? Is that a timestamp?

dubeejw: I see two different possible values for dataChanged.

screen shot 2017-10-17 at 10 33 53 am

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
dubeejw: I understand that the trigger will not be fetched on the read based on how the OpenWhisk infrastructure is setup. However, is there a test to make sure this provider errors as expected?

action/kafkaFeedWeb.js
dubeejw: Can you open an issue for this?

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: no, there is no absolute guarantee of the namespace when this runs

action/kafkaFeedWeb.js
abaruni: https://github.com/apache/incubator-openwhisk-package-kafka/issues/218

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: https://github.com/apache/incubator-openwhisk-package-kafka/issues/219

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
abaruni: time in millis

PR #224, opened 2017-11-16
add parameters as query parameters for get request

abaruni:
fix for `trigger get` failing in production because of executing a get request with a body


abaruni:
@csantanapr @jasonpet

PR #226, opened 2017-11-16
move positive trigger get test to health class

abaruni:


abaruni:
@csantanapr can you have a look

PR #229, opened 2017-11-30
verify trigger fire responds with any successful status code

abaruni:


csantanapr:
@jasonpet could you review this PR

provider/consumer.py
```diff
@@ -353,8 +353,12 @@ def __fireTrigger(self, messages):
 
                     # Manually commit offset if the trigger was fired successfully. Retry firing the trigger
                     # for a select set of status codes
-                    if status_code == 200:
-                        logging.info("[{}] Fired trigger with activation {}".format(self.trigger, response.json()['activationId']))
+                    if status_code in range(200, 300):
```
csantanapr: that might be 299 not 300 if range is inclusive

abaruni: the lower bound is inclusive while the upper bound is exclusive. i checked 😉

csantanapr: haha !

PR #230, opened 2017-12-06
Add rule and action to trigger tests which rely upon activation records

abaruni:
In order to support the removal of activation records for any trigger fires not attached to a rule, the tests which rely upon the presence of trigger activation records have been updated to create a simple hello action and a rule


abaruni:
@jasonpet will you give this a look?

abaruni:
@csantanapr fyi

csantanapr:
@jasonpet could you review and let me know

jasonpet:
LGTM

markusthoemmes:
What's the intended change in behavior here? Side-effects of `hello.js` are not observable so tests till evaluate activation list, no?

PR #231, opened 2017-12-12
make number of retries configurable via system property

abaruni:
when polling for activation records


jasonpet:
LGTM

abaruni:
@csantanapr will you have a look and merge?

PR #236, opened 2018-01-26
disable consumer after 5 failed attempts to connect to kafka brokers

abaruni:


csantanapr:
@dubeejw can you review 👁

PR #238, opened 2018-01-29
Adjust threshold of authentication errors with brokers before disabling consumer

abaruni:


abaruni:
@dubeejw @csantanapr would you mind giving this some 👀

provider/consumer.py
```diff
@@ -133,6 +133,18 @@ def __init__(self, trigger, params, sharedDictionary):
 
         self.authErrors = 0
 
+        # We want to account for the number of brokers when deciding the maximum number
+        # number of auth errors to allow
+        self.maxAuthErrors = len(self.brokers) * 30
```
abaruni: is this enough? should it be 50 attempts per broker? 100?

PR #241, opened 2018-01-31
Allow producer to batch messages

dubee:
POC of producer sending batched messages when an array is provided for the `value` field.

Needs tests.

### DIFF
--- a/action/messageHubProduce.py
+++ b/action/messageHubProduce.py
@@ -83,23 +83,20 @@ def main(params):
     # we successfully connected and found the topic metadata... let's send!
     if producer is not None:
         try:
-            logging.info("Producing message")
+            value = validatedParams['value']
+            batch = isinstance(value, list)
 
             # only use the key parameter if it is present
-            value = validatedParams['value']
             if 'key' in validatedParams:
                 messageKey = validatedParams['key']
-                future = producer.send(
-                    topic, bytes(value, 'utf-8'), key=bytes(messageKey, 'utf-8'))
             else:
-                future = producer.send(topic, bytes(value, 'utf-8'))
+                messageKey = None
 
-            # future should wait all of the remaining time
-            future_time_seconds = math.floor(getRemainingTime())
-            sent = future.get(timeout=future_time_seconds)
-            msg = "Successfully sent message to {}:{} at offset {}".format(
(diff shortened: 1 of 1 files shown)


dubeejw:
Could maybe require a `values` parameter instead of `value` for batched messages.

action/messageHubProduce.py
```diff
@@ -108,6 +105,42 @@ def main(params):
...
+    logging.info("Producing batched messages")
+
+    if messageKey is not None:
+        for val in values:
+            producer.send(topic, bytes(val, 'utf-8'), key=bytes(messageKey, 'utf-8'))
+    else:
+        for val in values:
+            producer.send(topic, bytes(val, 'utf-8'))
+
+    producer.flush()
```
dubeejw: Cannot get partition and offset now for log statement?

dubeejw: Doc for `flush()`: http://kafka-python.readthedocs.io/en/master/apidoc/KafkaProducer.html#kafka.KafkaProducer.flush.

dubeejw:
Test:
  - Produce batch messages
    - Poll activations and search for the messages sent in the batch
      - Batched messages may be in one or more activation record

action/messageHubProduce.py
```diff
@@ -83,23 +83,20 @@ def main(params):
     # we successfully connected and found the topic metadata... let's send!
     if producer is not None:
         try:
-            logging.info("Producing message")
-
             # only use the key parameter if it is present
             value = validatedParams['value']
+            batch = isinstance(value, list)
```
dubeejw: Could maybe require a values parameter instead of value for batched messages.

dubeejw: Would need to error if arguments are passed to both parameters in that case though.

action/messageHubProduce.py
```diff
@@ -83,23 +83,20 @@ def main(params):
     # we successfully connected and found the topic metadata... let's send!
     if producer is not None:
         try:
-            logging.info("Producing message")
+            value = validatedParams['value']
```
dubeejw: Could maybe require a `values` parameter instead of `value` for batched messages. Would need to error if arguments are passed for both parameters though.

rabbah:
closing as abandoned.

PR #243, opened 2018-01-31
remove assumption that partitions will be non-empty when passed into callbacks

abaruni:
i made a very bad assumption that the list of TopicPartitions passed into the assign and revoke callbacks would be nonempty. this is causing uncaught exception and consumer restart. i've removed this assumption. didn't need it anyway. it was used just for logging, but there was no useful information contained within


abaruni:
@dubeejw could you have a look

PR #244, opened 2018-01-31
Fix tests that fire trigger when message is consumed

dubeejw:
Tests poll for 100 activations matching `triggerName` for `maxRetries`. Doing this means we have to worry about the clock for the backend being in sync with the machine the health test runs on. Instead we can just get the last `x` related activations in one request and repeat the test `maxRetries` until we get the activation ID we are looking for.

Also we may have more than one matching activation due to the retry loop.

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -19,14 +19,17 @@ package system.health
 
 import java.time.Clock
 import java.time.Instant
+import java.time.temporal.ChronoUnit
 
 import system.utils.KafkaUtils
 
 import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
+
 import org.junit.runner.RunWith
 import org.scalatest.{BeforeAndAfterAll, FlatSpec, Inside, Matchers}
 import org.scalatest.junit.JUnitRunner
+
 import common.JsHelpers
 import common.TestHelpers
 import common.TestUtils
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -157,7 +157,6 @@ class BasicHealthTest
       }
 
       retry({
-        val start = Instant.now(Clock.systemUTC())
```
dubeejw: Why worry about the clocks being in sync?

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -173,8 +169,8 @@ class BasicHealthTest
           _.response.success shouldBe true
         }
 
-        println("Polling for activations")
-        val activations = wsk.activation.pollFor(N = 100, Some(triggerName), since = Some(start), retries = maxRetries)
+        println(s"Getting last 200 activations for $triggerName")
```
dubeejw: Get last 10 instead?

dubeejw:
@jberstler, any thoughts on this?

tests/src/test/scala/system/health/BasicHealthTest.scala
dubeejw: Pause a few seconds here just to let the activation get posted in the DB.

dubeejw:
Ah, I see. We are spamming the server with `activation list` requests with `pollFor` when we can just do a timeout based from `60 * maxRetries` then do one `activation list`.

PR #246, opened 2018-02-01
renew client connection when token expires

abaruni:
http://python-cloudant.readthedocs.io/en/latest/client.html#cloudant.client.CouchDB


abaruni:
@dubeejw could you have a look, please?

PR #247, opened 2018-02-07
Fix uncaught exception occurring when reading in a message containing non-unicode bytes

abaruni:
also, some doc updates and test updates


abaruni:
could not find a way to effectively create a scala test for this. i can test manually and verify the change works. i've attempted to create a python action to write non-unicode message to kafka. problem is that the non-unicode bytes get escaped when reading in the file while creating the action. manually invoking the action works. but i don't see away around the encoding and escaping/replacing that java/scala do

abaruni:
@dubeejw would you mind having a look

provider/consumer.py
```diff
@@ -446,6 +446,13 @@ def __getOffsetList(self, messages):
         return offsets
 
     def __encodeMessageIfNeeded(self, value):
+        # let's make sure whatever data we're getting is utf-8 encoded
+        try:
+            value = value.encode('utf-8')
+        except UnicodeDecodeError:
+            logging.warn('[{}] Value contains non-unicode bytes. Replacing invalid bytes.'.format(self.trigger))
+            value = unicode(value, errors='replace').encode('utf-8')
```
dubeejw: May want to doc this behavior in the future.

provider/consumer.py
```diff
@@ -454,6 +461,7 @@ def __encodeMessageIfNeeded(self, value):
             except ValueError:
                 # no big deal, just return the original value
                 logging.warn('[{}] I was asked to encode a message as JSON, but I failed.'.format(self.trigger))
+                value = "\"{}\"".format(value)
```
dubeejw: May want to doc this too.

README.md
```diff
@@ -232,6 +232,80 @@ Here is an example of a batched trigger payload (please note the change in the *
  }
  ```
 
+### Checking the status and configuration of a trigger
```
dubeejw: Very informative doc updates. :)

PR #250, opened 2018-02-21
Remove error callback

abaruni:


abaruni:
@dubeejw will you have a look when you've got time? 

fyi @csantanapr

PR #255, opened 2018-03-05
check for 204 before attempting to fetch response json

abaruni:


abaruni:
this is a fix needed for the removal of activation records for triggers that aren't bound to an action via a rule. when fetching the response json for the trigger fire, the python-requests library will raise an exception when there is no content. added a check for 204 explicitly and log a successful fire

PR #256, opened 2018-03-05
perform credentials check before creating consumer

abaruni:


abaruni:
there are two changes here:

1) Perform a check of the messagehub credentials (if it is messagehub) and topic. This is to ensure that the authkey and topic provide when the trigger was originally created are still valid. there is the possibility that credentials and/or a topic could be deleted without the trigger being deleted. this leaves around zombie consumers who never stop attempting to connect to a broker. the idea here is to do the check when creating a consumer for an active trigger. if the authkey and topic are valid, then we will create the consumer as usual. if either are invalid. then the consumer will not be created and the trigger will be disabled in the db. if encounter any errors while trying to validate, then we'll retry. if at the end of the alloted retries, we don't succeed in getting either a 200 or 403, then we'll treat it as though we did get a 200. the motivation being that we'll eventually catch it on subsequent deploys. better to err on the side of caution than to mistakenly disable a valid trigger

2) do not create consumers for disabled triggers

provider/service.py
```diff
@@ -106,7 +113,7 @@ def run(self):
                                     # running trigger should become disabled
                                     # this should be done regardless of which worker the document claims to be assigned to
                                     logging.info('[{}] Existing running trigger should become disabled'.format(change["id"]))
-                                    existingConsumer.disable()
+                                    existingConsumer.shutdown()
```
dubeejw: Need to update doc above that mentions that the consumer should be disabled.

provider/service.py
```diff
@@ -145,17 +152,48 @@ def stopChangesFeed(self):
     def createAndRunConsumer(self, doc):
         triggerFQN = doc['_id']
 
+        if doc['isMessageHub']:
+            retries = 0
+            max_retries = 5
+            url = '{}/admin/topics'.format(doc['kafka_admin_url'])
```
dubeejw: Will the admin URL throttle requests?

provider/service.py
```diff
@@ -145,17 +152,48 @@ def stopChangesFeed(self):
...
         # Creating this object is lightweight and does not initialize any connections
         consumer = Consumer(triggerFQN, doc)
         self.consumers.addConsumerForTrigger(triggerFQN, consumer)
 
-        if self.__isTriggerDocActive(doc):
-            logging.info('[{}] Trigger was determined to be active, starting...'.format(triggerFQN))
-            consumer.start()
-        else:
-            logging.info('[{}] Trigger was determined to be disabled, not starting...'.format(triggerFQN))
+        logging.info('[{}] Starting consumer...'.format(triggerFQN))
```
dubeejw: Do these changes affect the other occurrence of the call to `__isTriggerDocActive` in this file?

provider/service.py
```diff
@@ -145,17 +152,48 @@ def stopChangesFeed(self):
...
 
+        if doc['isMessageHub']:
+            retries = 0
+            max_retries = 5
+            url = '{}/admin/topics'.format(doc['kafka_admin_url'])
+            username = doc['username']
+            password = doc['password']
+            topic = doc['topic']
+
+            while retries < max_retries:
```
dubeejw: Seems like we could have a test here for invalid auth.

provider/service.py
```diff
@@ -145,17 +152,48 @@ def stopChangesFeed(self):
...
+                    reason = 'Invalid authKey. {} returned 403 using authKey {}{}.'.format(url, username, password)
+                    self.database.disableTrigger(triggerFQN, 403, reason)
+                    return
+                else:
+                    sleepyTime = pow(2, retries)
+                    logging.info('Received status code {} for {} while validating authKey. Retrying in {} second(s)'.format(status_code, triggerFQN, sleepyTime))
+                    time.sleep(sleepyTime)
+                    retries += 1
+
         # Create a representation for this trigger, even if it is disabled
```
abaruni: i need to update this comment as well

provider/service.py
abaruni: yes. currently we detect a change in the changes feed and check whether that consumer is in our in-memory cache. this is still the case, however, since we will not be creating disabled consumers. the only time we'll hit the code path where we are going from inactive -> active **and**

abaruni: we **do** have that consumer in memory is when we disable a trigger for reasons like client errors. doing that is something that may likely change in the future. so that whole logic would need reworking

provider/service.py
abaruni: yeah i'm afraid i'm going to have to change a doc in cloudant for that. need to get that figured out

dgrove-oss:
Closing as abandoned

PR #257, opened 2018-03-06
fail to create feed trigger on explicit 403

abaruni:


abaruni:
the error returned when checking messagehub credentials may not necessarily be due to a 403. this add an explicit check for 403. if the error is not a 403, then continue on as if though we'd received a 200 and let the provider disable the trigger if, when it does its check before creating the consumer, it fails with 403

PR #262, opened 2018-04-11
additional health field to monitor failed trigger fires

abaruni:
additional field in health endpoint keeping count of failed trigger fires. incremented after receiving client errors or after exhausting max number of retries


rabbah:
Closing as abandoned.

PR #264, opened 2018-04-17
remove use of messageHubProduce action in health test and use java producer

abaruni:


tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -163,15 +166,21 @@ class BasicHealthTest
...
+      val future = producer.send(record)
+
+      producer.flush()
+      producer.close()
+
+      try {
+        val result = future.get(60, TimeUnit.SECONDS)
+
+        println(s"Produced record to topic: ${result.topic()} on partition: ${result.partition()} at offset: ${result.offset()} with key: $key and value: $currentTime.")
+      } catch {
```
dubee: Should we retry the send if an exception occurs?

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -163,15 +166,21 @@ class BasicHealthTest
...
-        "key" -> key.toJson,
-        "value" -> currentTime.toJson
-      ))) {
-        _.response.success shouldBe true
+      val producer = kafkaUtils.createProducer()
+      val record = new ProducerRecord(topic, key, currentTime)
+      val future = producer.send(record)
+
+      producer.flush()
+      producer.close()
```
dubee: Is it safe to close the producer before we are guaranteed that the message was sent?

tests/src/test/scala/system/health/BasicHealthTest.scala
abaruni: i don't think it's necessary to retry at this time. if we do encounter failed sending in the future then it is something we should revisit and likely implement. this has less of a chance of failure compared to the produce action

Issue #269, opened 2018-06-07
Binary Encoding causes Corruption


maximann:
It looks to me like https://github.com/apache/incubator-openwhisk-package-kafka/blob/449bbae13e813ba4dcd11dc33f47ab29d5e3541a/provider/consumer.py#L455 encodes all values as UTF-8 before base64. My understanding is that some byte values are not valid UTF-8, thus corruption can occur at the line above. ISO-8859-1 seems like the better choice. I'd be happy to file a bug if this seems like a reasonable hypothesis and if someone could point me to where to do that. I was also wondering if someone has encountered a similar issue before, especially since binary serializers such as Avro are popular for Kafka.

Slack Transcript:
id you try the with the parameter “isBinaryValue” set to true ?

Max Manndorff [18 hours ago]
Yes

csantanapr [18 hours ago]
I believe the intention is to take the binary data encode base64 to be able to be sent to the controller http rest api

csantanapr [18 hours ago]
The controller will take the base64 in the json, passed it down thru kafka, invoker sends the base64 data in a http post request to the function runtime, the function receives the application/json with the this field base64 data

Max Manndorff [18 hours ago]
Right, that makes sense. But my understanding is all of this happens after binary data was consumed in the provider code linked above, so the corrupted data gets passed along just fine and makes it to the invocation correctly. But when binary messages are consumed from the external kafka cluster (the cluster which the trigger is setup for), and encoded as a UTF-8 strings before being passed along to the controller, the corruption has already happened. Or at least that's what I think is happening.

csantanapr [17 hours ago]
I think we saw a problem in practice we found messages had corrupted data and the line 456 is what tries to attempt to clean

Max Manndorff [17 hours ago]
Definitely seeing that in my logs:
```[2018-06-06T23:24:13.040Z] [WARNING] [??] [kafkatriggers] [/core/qmpTrigger] Value contains non-unicode bytes. Replacing invalid bytes.```

Max Manndorff [17 hours ago]
If you're ok with it I'll try to fix this for binary formats such as Avro tomorrow and send a diff your way.

csantanapr [17 hours ago]
You mean don’t do encode utf-8 when isBinaryValue=true ?

csantanapr [17 hours ago]
My concern is current users how are they impacted in production with the change

csantanapr [17 hours ago]
When they proceess the input value in the function and they decode base64 they get UTF-8 data and work on that data. (edited)

csantanapr [17 hours ago]
[...]

PR #270, opened 2018-06-07
Fix binary encoding

maximann:
This addresses #269 
Note: I'm not a python coder, so I'm sure there might be better ways to achieve this. 

When encoding binary data (ascii string) as utf with variable length encoding only 7 bits are preserved. The 8th bit has a special meaning to indicate the variable encoding "continuation bit". This will of course corrupt any true binary data that has values larger than 127.

A secondary issue addressed in this MR relates to the base 64 encoding mechanism. The method called previously inserts newline characters in the encoded string every 76 characters, something that's not typically expected in newer encoding libraries. I've added a new flag which allows encoding without newline characters (which I assume is what most people will expect).

Finally, each message is encoded twice in the current implementation, once to retrieve the size and a second time to actually trigger the function. 

This fix has the potential to break existing functions that may rely on the old behavior. I'm not sure how to address that and would appreciate feedback. A new parameter to trigger the fixed binary encoding may be necessary.


dubee:
@abaruni, FYI.

dgrove-oss:
closing as abandoned

PR #272, opened 2018-06-27
Verify trigger fire by sideeffect.

cbickel:
Until now, we verify the kafkaprovider, by looking if the action in OpenWhisk has been executed with `activation list`. The problem with this call is, that the activation may not be in the list in time, because it only returns the result of an CouchDB view. If there is much load on the database, the view-computation may be behind and not return the activation.
As side effect, we use a trigger creation. The name of the trigger will be read from the kafka message. All other things (like credentials, ...) are already present in the action, that is invoked anyway.
To check if the trigger exists, we use the ID of the trigger. So no view is involved anymore.

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -19,25 +19,20 @@ package system.health
 
 import java.util.concurrent.{TimeUnit, TimeoutException}
 
+import com.jayway.restassured.RestAssured
+import common.TestUtils.NOT_FOUND
+import common._
+import org.apache.kafka.clients.producer.ProducerRecord
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.{BeforeAndAfterAll, FlatSpec, Inside, Matchers}
+import spray.json.DefaultJsonProtocol._
+import spray.json._
 import system.utils.KafkaUtils
+import whisk.utils.retry
 
 import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
-import org.junit.runner.RunWith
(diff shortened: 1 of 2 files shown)


jasonpet:
@abaruni - can you please review this?  I can merge once you approve it.

tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -212,11 +218,10 @@ class BasicHealthTest
...
       }
 
-      retry({
-        println("Polling for activations")
-        val activations = wsk.activation.pollFor(N = 1, Some(triggerName), retries = maxRetries)
-        assert(activations.nonEmpty)
-      }, N = 3)
+      // Check if the trigger, that should have been created as reaction on the kafka-message, has been created.
+      // The trigger should have been created by the action, that has been triggered by the kafka message.
+      // If we cannot find it, the most probably the action did not run.
```
mdeuser: maybe also add comments explaining why looking for the trigger's and/or rule action activation was problematic

csantanapr:
@cbickel This is very cool solution.
At one point I discussed with @jasonpet to have the action to write to another DB and check the entry there.
But this way we re-use the whisk DB with an entity, it can be a trigger, a package, or action.

@mdeuser 
The current verification is nondeterministic and problematic because couchdb views are eventually consistent, and there could be large delay until the view is updated and the activation record to show up.

tests/src/test/scala/system/health/BasicHealthTest.scala
abaruni: i don't think this is a good idea as this comment would just be cluttering the code documenting our prior experiences instead of explaining the behavior of the code or the logic behind the implementation, etc.

PR #274, opened 2018-07-16
Remove required access to the whole couchdb database.

cbickel:
Currently the kafka-trigger needs admin access on the whole couch-database, to get all databases that exist. If not all permissions are granted, the provider fails to start.
With this change, the access on all databases is not needed anymore, but only on the required database.


jasonpet:
@abaruni - please review and verify that this PR will not cause any issues.

PR #278, opened 2018-08-21
Ignore certificates on creating a trigger within the side-effect action.

cbickel:
If the test to create an entity as side-effect within the triggered action runs in an environment with self signed certificates, this test will fail.

This PR fixes this bug.

@abaruni Could you please have a look on this PR?


abaruni:
@jasonpet or @dubee would you merge when you can?

PR #286, opened 2018-09-19
Fix for overriding api host

abaruni:
fix the way we are handling trigger urls when env var overrides api host. also remove basic auth from trigger url

### DIFF
--- a/provider/consumer.py
+++ b/provider/consumer.py
@@ -213,18 +213,29 @@ class ConsumerProcess (Process):
         return secondsSince(self.lastPoll())
 
     def __triggerURL(self, originalURL):
+        parsed = urlparse(originalURL)
         apiHost = os.getenv('API_HOST')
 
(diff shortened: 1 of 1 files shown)


provider/consumer.py
```diff
@@ -213,18 +213,29 @@ def secondsSinceLastPoll(self):
...
+            # remove https://user:pass@host from url and replace it with just https://host
+            # we do this because we no longer need the basic auth in the url itself.
+            # we rely upon the HTTPBasicAuth handler or the IAMAuth handler
+            parts = parsed.netloc.split('@')
+
+            if len(parts) == 2:
+                host = parts[1]
+            else:
+                host = parts[0]
+
```
abaruni: we want to maintain backward compatibility with trigger docs that have the user:pass in the url itself. but since we are using an auth handler we want to remove it from the url if it has it

PR #287, opened 2018-09-26
Update MessageHubProduceTests to not rely on activation polling

dubee:
Removes activation polling from the MessageHubProduceTests to reduce intermittent failures.


dubee:
@jasonpet, update tests are passing. Can you have a look?

PR #291, opened 2018-09-28
MessageHubFeedTests Resiliency Updates

dubee:
Removes activation polling from the MessageHubFeedTests to reduce intermittent failures.


dubee:
@jasonpet, was having some trouble with Travis, but it's working now. Test pass locally as well.

PR #292, opened 2018-10-04
Ensure test consumers exists before producing messages

dubee:
Refactoring changes that allow tests to make sure a consumer is running before producing messages to that consumer.

### DIFF
--- a/tests/src/test/scala/system/health/BasicHealthTest.scala
+++ b/tests/src/test/scala/system/health/BasicHealthTest.scala
@@ -59,7 +59,7 @@ class BasicHealthTest
 
   behavior of "Message Hub feed"
 
-  it should "create a consumer and fire a trigger when a message is posted to message hub" in withAssetCleaner(wskprops) {
+  it should "create a consumer and fire a trigger when a message is posted to messagehub" in withAssetCleaner(wskprops) {
     val currentTime = s"${System.currentTimeMillis}"
 
     (wp, assetHelper) =>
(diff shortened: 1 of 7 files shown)


tests/src/test/scala/system/health/BasicHealthTest.scala
```diff
@@ -55,125 +55,24 @@ class BasicHealthTest
...
-
-  val consumerInitTime = 10000 // ms
-
-  val kafkaUtils = new KafkaUtils
-
   val maxRetries = System.getProperty("max.retries", "60").toInt
 
   behavior of "Message Hub feed"
 
-  it should "create a new trigger" in withAssetCleaner(wskprops) {
```
dubee: Duplicate test, so deleting it. The next text in this suite tests the same functionality.

tests/src/test/scala/system/utils/KafkaUtils.scala
```diff
@@ -52,9 +58,75 @@ class KafkaUtils {
...
+                println("Giving the consumer a moment to get ready")
+                Thread.sleep(KafkaUtils.consumerInitTime)
+
+                val uuid = activation.response.result.get.fields.get("uuid").get.toString().replaceAll("\"", "")
+                consumerExists(uuid)
+        }
+    }
+
+
+    def consumerExists(uuid: String) = {
```
csantanapr: 👏 great way to check firwst the health endpoint to be sure consumer is ready and polling. 👍

PR #293, opened 2018-10-10
Update MessageHub parameter validation

dubee:
Allows Message Hub parameters to be extracted from a parameter starting with `__bx_creds.messagehub`.

### DIFF
--- a/action/messageHubFeedWeb.js
+++ b/action/messageHubFeedWeb.js
@@ -177,27 +177,15 @@ function validateParameters(rawParams) {
 
         validatedParams.isMessageHub = true;
 
-        if (rawParams.__bx_creds && rawParams.__bx_creds.messagehub) {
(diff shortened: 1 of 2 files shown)


action/messageHubFeedWeb.js
```diff
@@ -175,6 +175,38 @@ function validateParameters(rawParams) {
             }
         }
 
+        validatedParams.isMessageHub = true;
+
+        if (rawParams.__bx_creds && rawParams.__bx_creds.messagehub) {
```
csantanapr: remove the duplication, just check which params to use, replace with single
```
return validateMessageHubParameters(rawParams.__bx_creds && rawParams.__bx_creds.messagehub ? rawParams.__bx_creds.messagehub : rawParams)
.then(p => {
    validatedParams = Object.assign(validatedParams, p)
    resolve(validatedParams)
})
.catch(error => {
    reject(error);
    return;
})
```

PR #306, opened 2018-11-12
Update Tests to Wait for Producer to Finish

dubee:
Updates the `produceMessage` method to wait for the producer to finish before continue with tests.

### DIFF
--- a/tests/src/test/scala/system/packages/MessageHubFeedTests.scala
+++ b/tests/src/test/scala/system/packages/MessageHubFeedTests.scala
@@ -199,6 +199,7 @@ class MessageHubFeedTests
 
       wsk.trigger.get(verificationName, NOT_FOUND)
 
+      // The producer will generate an error as the payload size is too large for the MessageHub brokers
       a[ExecutionException] should be thrownBy produceMessage(topic, verificationName, generateMessage(s"${currentTime}", testPayloadSize))
       a[Exception] should be thrownBy retry(wsk.trigger.get(verificationName), 60, Some(1.second))
   }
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/system/utils/KafkaUtils.scala
```diff
@@ -176,7 +189,7 @@ object KafkaUtils {
         val security_protocol = ("security.protocol", "SASL_SSL");
         val keySerializer = ("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
         val valueSerializer = ("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
-
+        val maxRequestSize = ("max.request.size", "3000000");
```
dubee: The producer payload size is not big enough to handle the payload in test: `not fire a trigger for a single oversized message`. This went undetected previously as `producer.flush` was not called.

tests/src/test/scala/system/packages/MessageHubFeedTests.scala
```diff
@@ -197,9 +198,9 @@ class MessageHubFeedTests
       val verificationName = s"trigger-$currentTime"
 
       wsk.trigger.get(verificationName, NOT_FOUND)
-      println("Producing an oversized message")
-      produceMessage(topic, verificationName, generateMessage(s"${currentTime}", testPayloadSize))
 
+      // The producer will generate an error as the payload size is too large for the MessageHub brokers
+      a[ExecutionException] should be thrownBy produceMessage(topic, verificationName, generateMessage(s"${currentTime}", testPayloadSize))
```
dubee: This error was previously undetected as `producer.flush` was never called.

PR #308, opened 2018-12-03
Update Docker Base Image

dubee:


dubee:
Test pass and base image contains no security warnings. @jasonpet

PR #309, opened 2018-12-05
Remove Unnecessary Information from Health Endpoint

dubee:
Removes CPU data from the health endpoint leaving only consumer health data.


jasonpet:
This CPU data is still useful and could be refactored out into a separate endpoint

rabbah:
Closing as abandoned.

PR #313, opened 2019-01-07
Update Base Image

dubee:
Uses Python 2.7.15 and librdkafka 0.11.6 to avoid security issues existing in previously used dependancies.


dubee:
@jasonpet, tests passing locally.

PR #317, opened 2019-02-14
make kind on action install configurable

jasonpet:


installCatalog.sh
```diff
@@ -25,6 +25,7 @@ DB_NAME="${4}ow_kafka_triggers"
 APIHOST="$5"
 WORKERS="$6"
 INSTALL_PRODUCE_ACTION=${INSTALL_PRODUCE_ACTION:="true"}
+ACTION_RUNTIME_VERSION=${ACTION_RUNTIME_VERSION:="nodejs:6"}
```
mdeuser: maybe make the default parameter value be the default nodejs runtime instead of nodejs:6.  nodejs:6 will soon be out of service and this default will not be applicable then..

PR #330, opened 2019-04-05
Upgrade base Python version

dubee:
Use the newer Python 2 version.


dubee:
Change log here https://www.python.org/downloads/release/python-2716/.

PR #332, opened 2019-04-09
Prevent parsing floats as Infinity and -Infinity

dubee:
When reading a very large or very small float from Kafka, the consumer may convert that number into `Infinity` or `-Infinity` respectively. These changes cause an exception to occur in the `json.loads` function to prevent `Infinity` or `-Infinity` values as the OpenWhisk controller will reject such JSON as being invalid.


provider/consumer.py
```diff
@@ -526,3 +526,14 @@ def __on_revoke(self, consumer, partitions):
 
     def __errorOnJSONConstant(self, data):
     	raise(ValueError('Invalid JSON detected.'))
```
mdeuser: how about providing a more detailed exception message similar to the ones from `__parseFloat`

dubee: @mdeuser, done.

PR #341, opened 2019-05-14
Test resiliency updates

dubee:
Always ensure a consumer is running after updating that consumer.

### DIFF
--- a/tests/src/test/scala/system/utils/KafkaUtils.scala
+++ b/tests/src/test/scala/system/utils/KafkaUtils.scala
@@ -34,10 +34,10 @@ import system.packages.ActionHelper._
 import org.apache.openwhisk.utils.JsHelpers
 
 import scala.concurrent.duration.DurationInt
-import scala.language.postfixOps
 import common.TestHelpers
 import common.TestUtils
 import common.WskTestHelpers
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/system/utils/KafkaUtils.scala
```diff
@@ -66,26 +66,27 @@ trait KafkaUtils extends TestHelpers with WskTestHelpers {
...
-    def createTrigger(assetHelper: AssetCleaner, name: String, parameters: Map[String, spray.json.JsValue]) = {
+    def createTrigger(assetHelper: AssetCleaner, name: String, parameters: Map[String, spray.json.JsValue]): String = {
         println(s"Creating trigger $name")
 
         val feedCreationResult = assetHelper.withCleaner(wsk.trigger, name) {
             (trigger, _) =>
                 trigger.create(name, feed = Some(s"/whisk.system/messaging/messageHubFeed"), parameters = parameters)
         }
 
-        withActivation(wsk.activation, feedCreationResult, initialWait = 5 seconds, totalWait = 60 seconds) {
```
dubee: Don't need to query for the activation as we already have it from trigger create.

PR #356, opened 2019-11-11
Do not update last canary everytime a database connection attempt occurs

dubee:
When the provider continuously encounters network errors from the database, the last canary time is always being set to the current time. This logic prevents monitoring from detecting that the canary has actually gone stale because of the network issues. Instead, only update the canary outside of the database reconnection loop.

### DIFF
--- a/provider/service.py
+++ b/provider/service.py
@@ -50,6 +50,7 @@ class Service (Thread):
 
     def run(self):
         self.canaryGenerator.start()
+        self.lastCanaryTime = datetime.now()
 
         while True:
             try:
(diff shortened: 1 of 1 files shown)


dubee:
To further clarify, the logs show the sequence below repeatedly. While the changes feed appeared to re-establish its connection, read timeouts occur when trying to read from the changes feed. This explains why an exception was not thrown before `self.lastCanaryTime = datetime.now()` was executed.
```
[kafkatriggers] [canary] Exception caught from changes feed. Restarting changes feed... 
[kafkatriggers] HTTPSConnectionPool(host='OMITTED', port=443): Read timed out. (read timeout=30) 
[kafkatriggers] Shutting down existing DB client 
[kafkatriggers] Starting new HTTPS connection (2): OMITTED
[kafkatriggers] Starting changes feed 
[kafkatriggers] Starting new HTTPS connection (1): OMITTED
[kafkatriggers] Database exists - connecting to it. 
```

PR #359, opened 2019-11-18
Allow feed to be deleted if trigger does not exist

dubee:
If for some reason a trigger has been deleted though the associated feed still exists, a user cannot delete the corresponding feed manually.

### DIFF
--- a/action/lib/common.js
+++ b/action/lib/common.js
@@ -38,7 +38,7 @@ function getTriggerURL(endpoint, triggerName) {
     return url;
 }
 
-function verifyTriggerAuth(triggerURL, auth) {
+function verifyTriggerAuth(triggerURL, auth, rejectNotFound) {
     var options = {
         method: 'GET',
         url: triggerURL,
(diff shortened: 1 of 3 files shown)


action/lib/common.js
```diff
@@ -52,8 +52,12 @@ function verifyTriggerAuth(triggerURL, auth) {
 
     return request(options)
         .catch(err => {
-            console.log(`Trigger auth error: ${JSON.stringify(err)}`);
-            return Promise.reject({ authError: 'You are not authorized for this trigger.'});
+            if (err.statusCode && err.statusCode === 404 && !rejectNotFound) {
+                return Promise.resolve()
```
dubee: Is this promise needed or can we do `return true`?

dubee: Either way, I think.

rabbah:
@dubee does this affect other openwhisk feed providers? Can you mention this on the dev list 🙏

dubee:
@rabbah, sure. This and https://github.com/apache/openwhisk-package-kafka/pull/360 affect the other providers.

jasonpet:
this functionality has always existed for the other providers

dubee:
@rabbah, correction on that last comment. Alarms and Cloudant providers already allow for deletion of a feed that is triggerless. Changes here are to have the equivalent functionality for the Kafka provider. https://github.com/apache/openwhisk-package-kafka/pull/360 on the other hand does apply to the other providers.

PR #360, opened 2019-11-18
Update existing trigger feeds on create instead of failing

dubee:
If for some reason a trigger is deleted, but the corresponding feed is not, allow a trigger in the same namespace with the same trigger name to be recreated with the existing feed.

### DIFF
--- a/action/lib/Database.js
+++ b/action/lib/Database.js
@@ -58,7 +58,26 @@ module.exports = function(dbURL, dbName) {
         return new Promise((resolve, reject) => {
             this.db.insert(params, (err, result) => {
                 if(err) {
-                    reject(err);
+                    if(err.statusCode && err.statusCode === 409) {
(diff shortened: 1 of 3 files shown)


action/lib/Database.js
```diff
@@ -58,7 +58,16 @@ module.exports = function(dbURL, dbName) {
         return new Promise((resolve, reject) => {
             this.db.insert(params, (err, result) => {
                 if(err) {
-                    reject(err);
+                    if(err.statusCode && err.statusCode === 409) {
+                        this.getTrigger(params.triggerName)
+                        .then(() => this.disableTrigger(params.triggerName))
```
dubee: I don't understand why all this async stuff is going on when the entire action functionality is blocking anyway...

dubee:
@jasonpet, will definitely need your Node.js expertise on this one.

PR #362, opened 2019-11-22
Upgrade Provider from Python 2 to 3

dubee:
Python 2 will reach end of life soon and Python 2 packages will be removed PIP eventually. 

Integration tests pass, but encoding needs to be double checked. Python 3 encoding behaves differently. Suspecting problems when bytes are produced into Kafka and `isJSONData == true`.

**Do not merge. Will notify the dev list.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-FROM python:2.7.16
+FROM python:3.8.0
 
 RUN apt-get update && apt-get upgrade -y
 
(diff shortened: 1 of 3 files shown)


dubee:
@steven0711dong

Dockerfile
```diff
@@ -33,7 +33,7 @@ ENV CPLUS_INCLUDE_PATH /usr/local/include
 ENV LIBRARY_PATH /usr/local/lib
 ENV LD_LIBRARY_PATH /usr/local/lib
 
-RUN pip install gevent==1.1.2 flask==0.11.1 confluent-kafka==${LIBRDKAFKA_VERSION} \
+RUN pip install gevent==1.5a2 flask==0.11.1 confluent-kafka==${LIBRDKAFKA_VERSION} \
```
dubee: gevent 1.4 does not work with latest version of Python.

PR #364, opened 2020-01-02
Fix consumer dead state bug

dubee:
When an existing consumer is deleted and then quickly recreated, that consumer will get stuck in a dead state. This can happen when using wskdeploy with a manifest that undeploys triggers and then immediately recreates those triggers quickly.

### DIFF
--- a/provider/service.py
+++ b/provider/service.py
@@ -68,11 +68,6 @@ class Service (Thread):
                     # check whether or not the feed is capable of detecting canary
                     # documents
                     if change != None:
-                        # Record the sequence in case the changes feed needs to be
-                        # restarted. This way the new feed can pick up right where
-                        # the old one left off.
-                        self.lastSequence = change['seq']
-
                         if "deleted" in change and change["deleted"] == True:
                             logging.info('[changes] Found a delete')
                             consumer = self.consumers.getConsumerForTrigger(change['id'])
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/system/packages/MessageHubFeedTests.scala
```diff
@@ -113,6 +114,62 @@ class MessageHubFeedTests
     runActionWithExpectedResult(actionName, "dat/multipleValueTypes.json", expectedOutput, false)
   }
 
+  it should "create a trigger, delete that trigger, and quickly create it again with successful trigger fires" in withAssetCleaner(wskprops) {
```
dubee: Test fails when used against provider without the included updates.

provider/service.py
```diff
@@ -109,7 +104,20 @@ def run(self):
                                 elif triggerIsAssignedToMe:
                                     logging.info('[{}] Found a change to an existing trigger'.format(change["id"]))
 
-                                    if existingConsumer.desiredState() == Consumer.State.Disabled and self.__isTriggerDocActive(document):
+                                    if existingConsumer.desiredState() == Consumer.State.Dead and self.__isTriggerDocActive(document):
```
dubee: For background info, there is a nanny runs periodically that delete consumers that are in a dead state. Since the consumer is already in a dead state waiting for the nanny to close the process, a quick recreate of the consumer -- within a 2 second time period -- will be ignored.

provider/service.py
```diff
@@ -123,6 +131,11 @@ def run(self):
                             self.lastCanaryTime = datetime.now()
                         else:
                             logging.debug('[changes] Found a change for a non-trigger document')
+
+                        # Record the sequence in case the changes feed needs to be
+                        # restarted. This way the new feed can pick up right where
+                        # the old one left off.
+                        self.lastSequence = change['seq']
```
dubee: Moved this in case of an unexpected exception.

PR #369, opened 2020-05-23
Change nodejs:6 to default.

rabbah:
@dgrove-oss I looked through the node action source to inspect the `require`s and I updated the package.json templates accordingly.


dgrove-oss:
Thanks for doing this @rabbah!

Shariar076:
thanks @rabbah

PR #370, opened 2020-06-18
Handle URL params gently

NikolayStarodubtsev:
Previously url variable has strictly provided "https" protocol,
which totally broke feed if "https" was provided in endpoint variable.
Add new function based on regex to handle both types with or
without protocol.

### DIFF
--- a/action/lib/common.js
+++ b/action/lib/common.js
@@ -107,7 +107,7 @@ function massageParamsForWeb(rawParams) {
 function getWebActionURL(endpoint, actionName) {
     var apiHost = addHTTPS(endpoint);
 
-    return `${apiHost}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
+    return `${apiHost}/api/v1/web/whisk.system/messagingWeb/${actionName}`;
 }
 
 function createTrigger(endpoint, params, actionName) {
--- a/action/lib/common.js
+++ b/action/lib/common.js
@@ -26,14 +26,21 @@ function triggerComponents(triggerName) {
     };
 }
 
+function addHTTPS(url) {
+    if (!/^(?:f|ht)tps?\:\/\//.test(url)) {
+        url = "https://" + url;
+    }
+    return url;
+}
+
 function getTriggerURL(endpoint, triggerName) {
-    var massagedAPIHost = endpoint.replace(/https?:\/\/(.*)/, "$1");
+    var massagedAPIHost = addHTTPS(endpoint)
 
     var components = triggerComponents(triggerName);
     var namespace = components.namespace;
     var trigger = components.triggerName;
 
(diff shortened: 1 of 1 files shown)


action/lib/common.js
```diff
@@ -26,14 +26,21 @@ function triggerComponents(triggerName) {
     };
 }
 
+function addHTTPS(url) {
+    if (!/^(?:f|ht)tps?\:\/\//.test(url)) {
```
rabbah, line 30: i don't think `ftps` is relevant here - shouldn't this be only http or https?

action/lib/common.js
```diff
@@ -98,7 +105,9 @@ function massageParamsForWeb(rawParams) {
 }
 
 function getWebActionURL(endpoint, actionName) {
-    return `https://${endpoint}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
+    var massagedAPIHost = addHTTPS(endpoint);
+
+    return `${massagedAPIHost}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
```
rabbah, line 110:
```suggestion
    return `${apiHost}/api/v1/web/whisk.system/messagingWeb/${actionName}`;
```

action/lib/common.js
```diff
@@ -98,7 +105,9 @@ function massageParamsForWeb(rawParams) {
 }
 
 function getWebActionURL(endpoint, actionName) {
-    return `https://${endpoint}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
+    var massagedAPIHost = addHTTPS(endpoint);
```
rabbah, line 108:
```suggestion
    var apiHost = addHTTPS(endpoint);
```

rabbah:
Thanks @NikolayStarodubtsev for the contribution. I left some comments and suggestions.

action/lib/common.js
NikolayStarodubtsev: For sure. Will fix it =)

action/lib/common.js
NikolayStarodubtsev: Add this to last commit

action/lib/common.js
NikolayStarodubtsev: and this

action/lib/common.js
```diff
@@ -98,7 +105,9 @@ function massageParamsForWeb(rawParams) {
 }
 
 function getWebActionURL(endpoint, actionName) {
-    return `https://${endpoint}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
+    var apiHost = addHTTPS(endpoint);
+
+    return `${apiHost}/api/v1/web/whisk.system/messagingWeb/${actionName}.http`;
```
rabbah, line 110:
```suggestion
    return `${apiHost}/api/v1/web/whisk.system/messagingWeb/${actionName}`;
```

PR #379, opened 2021-05-04
improved logging in theDoctor  file 

groezing:
Improve the logging and if statement logic to detect the situation when the doctor does not restart a broken  consumer


groezing:
Accidentically opened in wrong repo

PR #385, opened 2022-05-25
Support dedicated providers

jiangpengcheng:
in our downstream, we provide OpenWhisk as a FaaS service, and now we want to provide kafka feeds features to users, we can use current code to run some kafka feed providers for users but it cannot satisfy below requirements:

1. users' kafka brokers may behind a firewall which block requests from kafka feed providers maintained by us
2. if user's kafka produce massive data, our common kafka feed providers may not be able to handle these data quickly

so this pr provides a way to make user can use their own feed providers

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,22 @@ Example:
 $ wsk trigger create MyKafkaTrigger -f /whisk.system/messaging/kafkaFeed -p brokers "[\"mykafkahost:9092\", \"mykafkahost:9093\"]" -p topic mytopic -p isJSONData true
 ```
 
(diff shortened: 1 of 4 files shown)


style95 (approved):

dgrove-oss (commented):
This looks plausible to me.  I did not test it (and I think travis testing for this repository is basically just a scanCode), but if it is working in your downstream it should be ok.

jiangpengcheng:
yes, I have tested this and it worked as expected

style95:
@dgrove-oss Thanks for the confirmation!

